PR #45, opened 2015-09-04
no signals in threads...

andre-merzky:


src/radical/utils/misc.py
```diff
@@ -232,6 +232,10 @@ class will install a signal handler for SIGUSR1.  When that signal is
 
     def __init__ (self) :
 
+        if 'MainThread' not in threading.current_thread().name:
```
marksantcroos: Is that a python name or something we set?

andre-merzky: That is Python's default name for the main thread.  I am not sure if that can be changed on purpose by an application though...

PR #50, opened 2015-09-17
Feature/signature comment

andre-merzky:
trivial PR.  Should those commits go directly to devel, maybe?


marksantcroos:
Comments can go to devel directly.

PR #53, opened 2015-09-19
fix setting of default log level, specifically for numeric settings

andre-merzky:


src/radical/utils/logger.py
```diff
@@ -109,10 +110,18 @@ def get_logger(name, target=None, level=None):
...
         level = os.environ.get('SAGA_VERBOSE', level).upper()
 
+    # translate numeric levels into symbolic ones
+    level = {50 : 'CRITICAL',
+             40 : 'ERROR',
+             30 : 'WARNING',
+             20 : 'INFO',
+             10 : 'DEBUG',
+              0 : _DEFAULT_LEVEL}.get(level, level)
+
```
marksantcroos: Given that we don't deal with any levels in between these numeric levels, whats the use of numeric level in the first place?

andre-merzky: This: radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/controller/pilot_launcher_worker.py#L618
`logging.getEffectiveLevel()` apparently returns an int, and that needs interpretation on the agent side then.

marksantcroos: Obviously we could change that line, but I guess what you want to say is that internally it only deals with numeric?

andre-merzky: Yes, correct.  I wasn't aware of this actually :/  But yes, thus I think it is better to also support those numbers.

Those are strange values though, I agree, one would think that the intermediate  values would also be usable, but I have at least not seen this documented...

PR #56, opened 2015-09-28
Feature/demo

andre-merzky:
This goes together with feature/demo of RP!

### DIFF
--- a/src/radical/utils/logger.py
+++ b/src/radical/utils/logger.py
@@ -210,13 +210,14 @@ def get_logger(name, target=None, level=None):
     # report, for example, demo output whereever we have a logger.
     def report(logger, style, msg):
         if logger._report:
-            if   style == 'title' : logger._reporter.title(msg)
-            elif style == 'header': logger._reporter.header(msg)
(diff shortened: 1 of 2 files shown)


src/radical/utils/logger.py
```diff
@@ -190,6 +206,27 @@ def get_logger(name, target=None, level=None):
...
+    # report, for example, demo output whereever we have a logger.
+    def report(logger, style, msg):
+        if logger._report:
+            if   style == 'title' : logger._reporter.title(msg)
+            elif style == 'header': logger._reporter.header(msg)
+            elif style == 'info'  : logger._reporter.info(msg)
+            elif style == 'ok'    : logger._reporter.ok(msg)
+            elif style == 'warn'  : logger._reporter.warn(msg)
+            elif style == 'error' : logger._reporter.error(msg)
+            else                  : logger._reporter.plain(msg)
```
marksantcroos: There is no `plain` method.

andre-merzky: Ah, right....  Thanks, will fix.

PR #57, opened 2015-09-29
default logging targe=t should be stderr

andre-merzky:


andre-merzky:
This is related to radical-cybertools/radical.pilot/issues/768

PR #58, opened 2015-10-08
Feature/demo

andre-merzky:
Lots of formatting tweaks, adding an idle mode for reporter, make LogReporter respect log level settings, simplify log reporter use (`logger.demo('info', msg) --> logger.report.info(msg)`)


andre-merzky:
I left the `logger.demo('info', msg)` syntax in place for backward compatibility -- but since this is young code, I would also not mind removing it.  Its is not used at the moment...

marksantcroos:
> I would also not mind removing it. Its is not used at the moment...

Im all for unified interfaces.

andre-merzky:
I'll do that then.  brb :)

andre-merzky:
I would not mind if that would be merged soon, for testing documentation examples etc.

marksantcroos:
I believe we didnt do this earlier because it broke RP devel IIRC.

marksantcroos:
Just verified, that is still the case:
```
2015-10-13 09:58:45,679: radical.pilot       : MainProcess                     : MainThread     : ERROR   : session create failed
  File "/Users/mark/.virtualenv/rp/lib/python2.7/site-packages/radical/pilot/session.py", line 137, in __init__
    logger.demo('info', 'create session %s' % self._uid)
AttributeError: 'Logger' object has no attribute 'demo'
```

andre-merzky:
Ah, right - but then again, that might be a decent point to start merging RP doc-things toward 0.37 by now.  Let me get those PRs in place then, if you don't mind...

marksantcroos:
sff-thanks-colour-tight

PR #61, opened 2015-10-27
allow reporter and debug log to coexist if log target is set.

andre-merzky:
Adresses https://github.com/radical-cybertools/radical.pilot/issues/879


marksantcroos:
Thanks!

Issue #64, opened 2015-11-06
Url.get_path() throws exception


mehdisadeghi:
Hi there,

According to this line of [documentation](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/url.py#L38) I should be able to do this:

    url = radical.utils.Url ("scheme://pass:user@host:123/path?query#fragment")

But surprisingly I get an error if I call `get_path` afterwards:

    >>> import radical.utils
    >>> url = radical.utils.Url ("scheme://pass:user@host:123/path?query#fragment")
    >>> url.get_path()
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/W5/sade/workspace/radical.utils/src/radical/utils/url.py", line 427, in get_path
        (path, query) = path
    ValueError: too many values to unpack
    >>> 

Apparently changing [this line](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/url.py#L427) might help:

     (path, query) = path -> (path, query) = path.split('?')

But as I digged ;-) deeper into saga code I realized there is an underlying `_urlobj` object which comes from [radical.utils.contrib.urlparse25](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/contrib/urlparse25.py) and it seems that saga is using its own url parsers according to some standards that I'm not familiar with. Therefore I don't know where should be changed to fix this behaviour. 

Anyway, here is an example of saga.Url which returns weired result:

    >>> from saga import Url
    >>> u2 = Url ("scheme://pass:user@host:123/path?query#fragment")
    >>> u2._urlobj
    ('scheme', 'pass:user@host:123', '/path?query#fragment', '', '', '')
    >>> u2.get_query()
    'query#fragment'
    >>> u2.get_fragment()
    ''
    >>> u2.get_username()
    'pass'
    >>> u2.get_password()
    'user'

I am not sure if I am using the correct string, but that is from previous example, and there is the same exception here for calling `u2.get_path()`.

I am just trying to pass some ssh options to the job service, specifically `"StrictHostKeyChecking=no"`.

andre-merzky:
Yes, you should be able to do this, this is  an error indeed.  

The underlying urlparse is quite a mess -- but IIRC, Ole could not find a better / working URL parser in python, so copied this one and adapted to our needs.  We have a long standing item to revisit that approach (see #30), but I have no idea when we'll get around actually doing so.

[...]

PR #66, opened 2015-11-06
Fix/url fragments

andre-merzky:


andre-merzky:
This PR is fixing #64 - the issue about limited scheme support as discussed in that ticket is not addressed here.  WIth this limited scope, the PR is ready to merge IMHO.

PR #71, opened 2016-03-05
scattered small fixes to logging

andre-merzky:
- don't barf on flush on closed/lost log handles
  - provide process of origin on stack traces
  - fix a default setting

### DIFF
--- a/src/radical/utils/reporter.py
+++ b/src/radical/utils/reporter.py
@@ -244,11 +244,17 @@ class Reporter(object):
             stream.write(msg)
             stream.write(self.COLORS['reset'])
             stream.write(self.COLOR_MODS['reset'])
-            stream.flush()
+            try:
+                stream.flush()
+            except Exception as e:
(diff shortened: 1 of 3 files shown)


src/radical/utils/reporter.py
```diff
@@ -244,11 +244,17 @@ def _out(self, color, msg):
             stream.write(msg)
             stream.write(self.COLORS['reset'])
             stream.write(self.COLOR_MODS['reset'])
-            stream.flush()
+            try:
+                stream.flush()
+            except Exception as e:
+                pass
```
marksantcroos: Why can the flush fail for various reasons but the write can't?

andre-merzky: I am not  sure, but I have seen this happening.   Doesn't make sense to me either :/

marksantcroos: Ok :)

PR #74, opened 2016-03-22
make sure log level is string before calling "upper()"

andre-merzky:
This will result in a hotfix release

### DIFF
--- a/src/radical/utils/logger.py
+++ b/src/radical/utils/logger.py
@@ -228,9 +228,6 @@ def get_logger(name, target=None, level=None):
     if env_name.startswith('RADICAL_SAGA'):
         level = os.environ.get('SAGA_VERBOSE', level)
 
-    # we want levels to be uppercase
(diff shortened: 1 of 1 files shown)


marksantcroos:
I was recently looking at this code, and wondering why we are duplicating the constants from the python logger. (The context was the guard for debug statements)

andre-merzky:
> I was recently looking at this code, and wondering why we are duplicating the constants from the python logger. (The context was the guard for debug statements)

Yeah, I don't like that either, but: we used to take the client side debug level from `logger.getEffectiveLevel()` or something like that, and put it into the agent config.  That call returns the numeric, not the string value, so we translated it back on the agent side.  That might not really be needed anymore, as we reduced the amount of magic we do on the level string by now, but I am not sure about that, would need to check.  So lets leave it in until the next time we stumble over it, and I'll try to clean it up then, ok?

PR #80, opened 2016-06-02
Fix/double release

andre-merzky:
duplicated freeing of shared resources indicates either a race condition in the lease manager, or a logic error in the user of the LM.  Either way, it actually happens at the point where the danger has *passed*, ie. after concurrent use of the leased channel, and we thus ignore this condition for now.


andre-merzky:
I should have mentioned that this popped up on some long-running RP experiments, from time to time.  I did not dig deep enough to know for sure who was the culprit in those cases...

PR #83, opened 2016-06-15
thread cancel should also terminate in the main thread

andre-merzky:


marksantcroos:
When/where is this used?

andre-merzky:
In component termination.  Termination requests can come from the main thread (in case of errors for example) of from other threads (like from the heartbeat callback).   To keep the behavior consistent, we always redirect such events to the main thread, which will need to except `KeyboardInterrupt` and `SystemExit` exceptions for this to work reliably.  When excepting these events we start the actual shutdown routines.

marksantcroos:
Why is the commit in a file named debug.py then?

andre-merzky:
Puh, good question.  Because I have no better place for it I guess, and it fits with all the threading and callstack stuff there.  Not much of an argument I guess... :P

marksantcroos:
Can you put it in a sane place? Does it need to part of RU in the first place?

andre-merzky:
Ok, let me find a better place.  But yes, I think utilities of that level should live in RU -- FWIW, there are some more I intent to move down from RP over time.

andre-merzky:
Done :)

marksantcroos:
Thanks!

andre-merzky:
also thanks :)

PR #85, opened 2016-07-24
Python 3 support.

ostrokach:
I ran the code-base through python-modernize and made a few other changes (e.g. adding `__next__` methods).

Required by radical-cybertools/saga-python#399.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,7 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+
+*.bak
+SDIST
+VERSION
--- a/src/radical/utils/algorithms.py
+++ b/src/radical/utils/algorithms.py
@@ -1,6 +1,9 @@
 
+from __future__ import absolute_import
+from __future__ import print_function
 import os
-import regex
+from . import regex
+from six.moves import range
 
 
 # ------------------------------------------------------------------------------
@@ -102,10 +105,10 @@ if __name__ == '__main__':
     pprint.pprint (test)
     pprint.pprint (collapse_ranges (test))
 
-    space = range(75)
+    space = list(range(75))
     parts = partition(space, 8)
     for part in parts:
-        print "%3d: %s" % (len(part), part)
+        print("%3d: %s" % (len(part), part))
 
 
 # ------------------------------------------------------------------------------
--- a/src/radical/utils/benchmark.py
+++ b/src/radical/utils/benchmark.py
@@ -1,4 +1,8 @@
 
+from __future__ import absolute_import
+from __future__ import print_function
+import six
+from six.moves import range
 __author__    = "Radical.Utils Development Team (Andre Merzky)"
 __copyright__ = "Copyright 2013, RADICAL@Rutgers"
 __license__   = "MIT"
@@ -12,8 +16,8 @@ import socket
 import threading
 import traceback
 
-import threads as rut
-import testing as rutest
+from . import threads as rut
+from . import testing as rutest
 
 
 # --------------------------------------------------------------------
@@ -26,7 +30,7 @@ class Benchmark (object) :
 
         try:
             import numpy
-        except ImportError, e:
+        except ImportError as e:
             raise RuntimeError ("Benchmark depends on numpy which is not installed")
     
         self.pre    = func_pre
@@ -36,12 +40,12 @@ class Benchmark (object) :
         self.lock   = rut.RLock ()
         self.events = dict()
     
-        if  isinstance (config, basestring) :
+        if  isinstance (config, six.string_types) :
             app_cfg = rutest.TestConfig (config)
         elif isinstance (config, dict) :
             app_cfg = config
         else :
-            print 'warning: no valid benchmarl config (neither filename nor dict) -- ignore'
+            print('warning: no valid benchmarl config (neither filename nor dict) -- ignore')
             app_cfg = dict()
 
         if  'benchmarks' in app_cfg :
--- a/src/radical/utils/contrib/urlparse25.py
+++ b/src/radical/utils/contrib/urlparse25.py
@@ -1,4 +1,6 @@
 
+from __future__ import absolute_import
+from __future__ import print_function
 __author__    = "Radical.Utils Development Team (Andre Merzky)"
 __copyright__ = "Copyright 2013, RADICAL@Rutgers"
 __license__   = "MIT"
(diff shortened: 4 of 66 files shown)


src/radical/utils/algorithms.py
```diff
@@ -1,6 +1,9 @@
 
+from __future__ import absolute_import
+from __future__ import print_function
 import os
-import regex
+from . import regex
+from six.moves import range
```
andre-merzky: I am not sure I understand why we would need `xrange` here (which I think is what six maps to?).  Let me look further if six is required elsewhere, otherwise I would be in favor of removing that dependency.  Am I missing something?

src/radical/utils/benchmark.py
```diff
@@ -36,12 +40,12 @@ def __init__ (self, config, name, func_pre, func_core, func_post) :
         self.lock   = rut.RLock ()
         self.events = dict()
     
-        if  isinstance (config, basestring) :
+        if  isinstance (config, six.string_types) :
```
andre-merzky: Ah :)

src/radical/utils/contrib/urlparse25.py
```diff
@@ -239,16 +241,18 @@ def urlsplit(url, scheme='', allow_fragments=True):
     _parse_cache[key] = v
     return v
 
-def urlunparse((scheme, netloc, url, params, query, fragment)):
+def urlunparse(xxx_todo_changeme):
```
andre-merzky: ?  I am not sure if this is a leftover?

src/radical/utils/debug.py
```diff
@@ -1,10 +1,12 @@
 
+from __future__ import absolute_import
+from __future__ import print_function
 import os
 import sys
 import time
 import pprint
 import signal
-import thread
+import six.moves._thread
```
andre-merzky: I don't see this being used - what is the effect of this?

src/radical/utils/misc.py
```diff
@@ -193,7 +195,7 @@ def time_diff (dt_abs, dt_stamp) :
 
     # make it easy to use seconds since epoch instead of datetime objects
     if  isinstance (delta, int)   or \
-        isinstance (delta, long)  or \
+        isinstance (delta, int)  or \
```
andre-merzky: same

src/radical/utils/read_json.py
```diff
@@ -58,10 +60,10 @@ def write_json (data,  filename) :
     # http://stackoverflow.com/questions/956867/#13105359
     def _byteify(input):
         if isinstance(input, dict):
-            return {_byteify(key):_byteify(value) for key,value in input.iteritems()}
+            return {_byteify(key):_byteify(value) for key,value in six.iteritems(input)}
```
andre-merzky: wouldn't `dict.iter()` work in cases like this?

src/radical/utils/read_json.py
```diff
@@ -58,10 +60,10 @@ def write_json (data,  filename) :
...
         elif isinstance(input, list):
             return [_byteify(element) for element in input]
-        elif isinstance(input, unicode):
+        elif isinstance(input, six.text_type):
             return input.encode('utf-8')
```
andre-merzky: Is there an explicit python3 check for unicode?  The encode should not be needed for basestring I would think.

src/radical/utils/read_json.py
```diff
@@ -119,7 +121,7 @@ def _unicode_to_strings (unicode_data) :
             ret.append (_unicode_to_strings (element))
         return ret
 
-    elif isinstance (unicode_data, unicode) :
+    elif isinstance (unicode_data, six.text_type) :
         return unicode_data.encode ('utf-8')
```
andre-merzky: same

src/radical/utils/reporter.py
```diff
@@ -1,40 +1,22 @@
...
-    #           'darkyellow'  : c.Style.DIM       + c.Fore.YELLOW  ,
-    #           'darkgreen'   : c.Style.DIM       + c.Fore.GREEN   ,
-    #           'darkblue'    : c.Style.DIM       + c.Fore.BLUE    ,
-    #           'darkcyan'    : c.Style.DIM       + c.Fore.CYAN    ,
-    #           'darkred'     : c.Style.DIM       + c.Fore.RED     ,
-    #           'darkmagenta' : c.Style.DIM       + c.Fore.MAGENTA ,
-    #           'darkblack'   : c.Style.DIM       + c.Fore.BLACK   ,
-    #           'reset'       : c.Style.RESET_ALL + c.Fore.RESET
-    #       }
-
```
andre-merzky: Can we leave that comment, please?  I wanted to eventually add a support for color palettes.  Not sure if I'll ever get around doing so, but... :)  Thanks!

src/radical/utils/threads.py
```diff
@@ -224,8 +225,8 @@ def is_main_thread():
 def cancel_main_thread():
 
     if not is_main_thread():
-        import thread
-        thread.interrupt_main()
+        import six.moves._thread
+        six.moves._thread.interrupt_main()
```
andre-merzky: I am somewhat worried about this one, and `_thread` in general.  Let me explain:

The python2 version of interrupt_main is broken, in that it conflicts with SIGINT signal handlers.  We work around by using different signals for interrup handling (which is cumbersome, but works). 

Further, python multiprocessing conflicts with threading in that several core libraries, specifically `logging`, potentially keep locked handles across forks, which leads to deadlocks.  We work around this by monkey-patching the logging module.

The Python support answer for both was always 'don't mix processes and threads!'.  While we have to accept that answer, that is precisely what we do, and I don't see any way around it either.  Have you seen, by any chance, any indication that Python3 behaves differently?  I this point I actually would prefer if it behaves the same, so that our patches remain functional.  But either way, that is something we'll need to do some testing on before getting it into production.

.gitignore
```diff
@@ -34,3 +34,7 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+
+*.bak
+SDIST
+VERSION
```
andre-merzky: `VERSION` actually should be in the repo.  SDIST should not appear in the source tree, only in the install tree.  Is that different for you?

andre-merzky:
Hey Alexey!  

many thanks for that effort!  

I could not find a reference to `__next__` in the diff -- what does this refer to?

andre-merzky:
Thanks again for this PR!  This is now dormant for quite a while though, and we are well under way to move toward Python3 by now, so I will close this PR in favor of those efforts.

Take care, Andre.

PR #86, opened 2016-07-24
Feature/raise on

andre-merzky:
```
    The purpose of this method is to artificially trigger error conditions for
    testing purposes, for example when handling the n'th unit, getting the n'th
    heartbeat signal, etc.  

    The tag parameter is interpreted as follows: on the `n`'th invocation of
    this method with any given `tag`, an exception is raised, and the counter
    for that tag is reset.
    
    The limit `n` is set via an environment variable `RU_RAISE_ON_<tag>`, with
    `tag` in upper casing.  The environment will only be inspected during the
    first invocation of the method with any given tag.  The tag counter is
    process-local, but is shared amongst threads of that process.
```

### DIFF
--- a/src/radical/utils/debug.py
+++ b/src/radical/utils/debug.py
@@ -258,43 +258,51 @@ def get_caller_name(skip=2):
 
 # ------------------------------------------------------------------------------
 #
-_raise_on_cnt = dict()
-def raise_on(tag, n, log=None):
+_raise_on_state = dict()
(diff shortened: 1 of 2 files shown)


andre-merzky:
the PR also contains a backward-compatible update to `cancel_main_thread()`, which now can use different signals.  This is essential toward the changed RP termination procedure.

PR #87, opened 2016-07-30
Feature/algs

andre-merzky:
This provides a number of algorithmic utilities, toward radical.analytics.  It also adds `get_hostip` and `get_hostname`, the former copied from RP.  This implies that netifaces is now a RU dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -189,7 +189,9 @@ setup_args = {
     'cmdclass'           : {
         'test'           : our_test,
     },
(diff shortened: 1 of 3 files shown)


marksantcroos:
Why is the analytics stuff not in the analytics repo? Is it really that generic?

andre-merzky:
Some of this might migrate into RA over time I guess.  But yes, this is indeed rather generic stuff, which for example does not depend on our profiling formats or specific event models.

PR #92, opened 2016-10-10
Feature/raise in thread

andre-merzky:
Work on threading and debug helpers

This also makes the `ru.Singleton` class safer to use over a `fork()` call.


marksantcroos:
This doesn't merge clean (anymore).

PR #93, opened 2016-11-15
Experiment/aimes

andre-merzky:
Bring devel back into sync with `experiment/aimes`.

the only major change here really is the addition of the radical-stack tool, which is supposed to make error reporting etc. somewhat simpler.


andre-merzky:
Thanks!  I'll restore the branch though, it's part of the stack for Matteo's experiments.

marksantcroos:
ACK.

PR #94, opened 2016-12-14
Feature/managed process

andre-merzky:
This Process class is a thin wrapper around multiprocessing.Process which specifically manages the process lifetime in a more cautious and cooperative way than the base class.

The class will need some updates on the code documentation, but otherwise now seems to do what it is supposed to do, and I would appreciate a code review.

Some basic tests are in `tests/unittests/test_process.py`, which specifically cover those cases relevant to RP.


andre-merzky:
Hi guys, 

this PR is at the heart of what we fighted against for the last 12 months (or more, really): clean termination of multiprocessed / multithreaded Python code.  The PR brings a wrapper around Python's `multiprocessing.Process` class, which attempts to manage process lifetime (including startup, initialization, failure detection and termination) *without* signals, cross-thread exceptions, or interrupts.  Unfortunately, all of the above are broken to one extent or the other in Python -- for details see the discussion on top of `src/radical/utils/process.py`, for demonstrators of the various cited Python tickets,  please check [here](https://github.com/radical-cybertools/radical.pilot/tree/feature/split_termination_4/docs/architecture/termination/issues).  

This PR might be slightly incomplete in some details (although it passes the respective tests already).  At this point, I am mostly interested in a review of the chosen approach etc, but since that code is so central to the stack, a thorough review would in fact be appreciated...

andre-merzky:
> In particular, the cancel() method in the Future class appears to require work in saga-python. Not sure if this has been done/will be/needs to be.

The corresponding SAGA branch would be `feature/futures`, which makes the switch from `ru.Task` to `ru.Future`.  The PR for that is here: https://github.com/radical-cybertools/saga-python/pull/608

For the `cancel()`: that is indeed not a problem on the SAGA layer, which right now is the only consumer of the `Future` class: the SAGA adaptors handle cancelation requests.  We could remove the `cancel()` call from the Future API for clarity.

andre-merzky:
> FIXMEs still exist in several places. I'm not really sure if these are fundamental to whether this branch can be considered 'working' and can be merged (in which case they can become TODOs and issue(s) need to be raised to track the future work to address them), or whether they need to be fixed now before merge.

Some of the FIXME's need to be addressed before merge, some can probably indeed be converged into TODO's.  I'll go through and sort them.  Either way: this needs indeed checking before merge.

andre-merzky:
> TODO in Process class re: fork/exec - what is the advantage here over the current implementation i.e. is it needed or is it gold-plating?

Python's multiprocessing class allows to do something like this:

```python
def work():
  while True: 
    print 'foo'
    time.sleep(1)

p = multiprocessing.Process(target=work)
p.start() 
```

This will do a `fork`, but will not exec a new executable, but instead jump to the given target method.

The problem with this is that several of Python's core libraries are actually not safe to be carried over a fork, *unless* an `exec` is called after that.  Good example: logging.  Assume a thread in the parent process logs a message while the `fork()` is called (in `p.start()`).  That thread will then hold a lock on the log stream.  The `fork()` will inherit that lock -- but will not inherit the thread, so the lock never gets released.  If our `work()` would call into the logging module, it would hang.

Specifically for logging we have some monkeypatching in place which releases all logs on `fork()`.  For other modules we can't do that, really, or don't know how to do it.  So we 'fork-and-pray' :-P  The clean solution would be to `exec` into a new process which then bootstraps to the same state we had, and then calls `work()` -- the `exec` would fundamentally provide a clean slate with respect to logs.

Summary: this would be great, but further increases code complexity :/

andre-merzky:
> FIXME in Process `__init__` - some places already check for _rup_spawned, how far do you want to go with these checks?

Yeah, this is about that check of `self._rup_spawned`.  I need to check for which methods this is required -- generally for all methods which look at the socket or the child process, as those only exist after spawning.

ibethune:
OK, some further comments/questions as I read through the whole smash in light of our discussion on Weds:

- Any reason why the initial value of _rup_started is False while other 'booleans' e.g. _rup_spawned etc. are None

- Can/should we control which threads can access the socketpair to avoid anything else interfering with the parent/child comms? e.g. the watcher thread is created with some attribute which we assert in _ru_msg_send() and _ru_msg_recv()

- Should we register for POLLPRI type events, it's not clear to me in what case data arriving would be of type POLLIN and when it might be POLLPRI (if so we also check for that in _rup_watch_socket())

- Probably the various messages that can be sent down the pipe e.g. 'terminating', 'work finished' etc. should not be hardcoded throughout, but rather defined at the top of the file, as per _ALIVE_MSG

- I might have missed this, but how does the watcher thread distinguish between a child exiting because it was done (and closing its end of the pipe) i.e. the parent carries on happily, vs. the child process died and therefore we need to tear everything down?

setup.py
```diff
@@ -269,8 +269,14 @@ def isgood(name):
...
     'cmdclass'           : {
         'test'           : our_test,
+<<<<<<< HEAD
+    },
+    'install_requires'   : ['future', 
+                            'colorama',
+=======
                            },
     'install_requires'   : ['colorama', 
+>>>>>>> e7b4a648ea5b99e8ab8e557eac500219742ce3d5
```
marksantcroos: FYI

andre-merzky:
Thanks Mark!

PR #95, opened 2017-01-22
rename ru.Thread to ru.Future

andre-merzky:
This PR copies the `ru.Thread` implementation to `ru.Future`.

`ru.Thread` has two main consumers: async ops in RS, and threading support in RP.  Async SAGA calls are semantically really Futures though, and this commit makes this explicit.  At the same time, it frees the `ru.Thread` implementation from Future related semantics which conflict with upcoming changes to thread lifetime management, as needed by RP.

This PR needs to be in sync with the respective merge of the RS branch `feature/future`.


andre-merzky:
PS.: I should note that *at this point* the change is backward compatible, as the original `ru.Thread` class is still present with the same API.  The branch has been merged forward into `feature/managed_process`, and backward incompatible changes (on `ru.Thread`) will happen only there.

PR #105, opened 2017-05-16
be friendly to MacOS

ibethune:


ibethune:
I checked and 'ps -efw' (i.e. without the --forest argument) gives sensible output on my Mac.  I think this can then be merged without further effort to recreate.

marksantcroos:
Thanks

PR #121, opened 2017-10-02
fixes issue 120 - incomplete profiles

### LINKED ISSUE #120: Sub-components not being shutdown properly
This is related to radical-cybertools/radical.pilot#1362

itomaldonado:
Fixes how the parent/child process relationship is terminated via sockets...

Increased the stop_timeout to 10 seconds.

Should fix #120

### DIFF
--- a/src/radical/utils/process.py
+++ b/src/radical/utils/process.py
@@ -28,7 +28,7 @@ _START_TIMEOUT = 5.0      # time to wait for process startup signal.
 _WATCH_TIMEOUT = 0.5      # time between thread and process health polls.
                           # health poll: check for recv, error and abort
                           # on the socketpair; is done in a watcher thread.
-_STOP_TIMEOUT  = 10.0     # time between temination signal and killing child
+_STOP_TIMEOUT  = 2.0     # time between temination signal and killing child
 _BUFSIZE       = 1024     # default buffer size for socket recvs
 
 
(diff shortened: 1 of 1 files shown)


src/radical/utils/process.py
```diff
@@ -28,7 +28,7 @@
 _WATCH_TIMEOUT = 0.5      # time between thread and process health polls.
                           # health poll: check for recv, error and abort
                           # on the socketpair; is done in a watcher thread.
-_STOP_TIMEOUT  = 0.1      # time between temination signal and killing child
+_STOP_TIMEOUT  = 10.0     # time between temination signal and killing child
```
andre-merzky: PBS gives us about 30 seconds in total from the first `SIGTERM` to `SIGKILL`, and we need to have finished termination by then.  10sec per process layer is cutting it tight, as we have 3 layers: (agent -> sub-agent -> component), plus some watcher-timeouts added on top.

This is a difficult balance :/

itomaldonado: I agree, I can suggest cutting it to a lower number, perhaps 2 seconds? 

The reason we have to increase this is simple, during shutdown, there are several waits that happen... at a simple glance this is what I see...

*0.5* from `_WATCH_TIMEOUT`
*0.1* from `events = self._ru_poller.poll(0.1)`
`msg = self._ru_msg_recv(_BUFSIZE)` is blocking --> which I suggest we change to something like *0.1*
all of which adds to ~ *0.7* seconds... and this is at the lowest level (i.e. at the radical.utils --> process.py level).

that does not take into account actual processing time for closing down the component itself (computation time).

So rounding up and doubling we get my suggestion of *2* seconds...

andre-merzky: Ack to that!  Its tricky to find the right balance in the given boundary conditions, but that looks sane to me. Thanks!

andre-merzky:
Thanks for digging into this ! :)

PR #124, opened 2017-10-19
added a handle-None check in the flush method of profile

itomaldonado:
simple `if not self._handle` check in the `flush()` method of the profiler, similar to what is done in the `prof()` method.


andre-merzky:
Bugger, did I really miss to merge this commit?  Dammit...
Sorry, and thanks!

PR #128, opened 2017-12-08
Feature/id namespaces

andre-merzky:
EnTK creates multiple IDs within a single session (for pipelines, stages, etc) - but those are potentially created in different processes.  This patch ensures that the RU ID semantics can be expanded across processes, by introducing a notion of namespaces which scopes the IDs.


andre-merzky:
thanks Vivek!

PR #129, opened 2018-01-16
Feature/examples debug

andre-merzky:
This adds very simple usage examples for some debug utilities.  It also switches from float based random specs for `raise_on` to int-based ones, which are easier IMHO.


src/radical/utils/debug.py
```diff
@@ -281,8 +281,13 @@ def get_caller_name(skip=2):
 
 # ------------------------------------------------------------------------------
 #
+_verb  = False
+if 'RADICAL_DEBUG' in os.environ :
+    _verb = True
```
iparask: There is an assumption here that whenever `RADICAL_DEBUG` exists in the environment it is going to be `True`. What if it is `False` or something else?

src/radical/utils/tracer.py
```diff
@@ -49,11 +49,13 @@ def _tracer (frame, event, arg) :
 
         filename = frame.f_globals["__file__"]
         lineno   = frame.f_lineno
+
```
iparask: I remember you did not like more than one empty lines?! :)

src/radical/utils/debug.py
andre-merzky: RADICAL_DEBUG can be empty or set to anything - it just needs to be present.  I wanted this to be a quick check (no additional string compares etc).

src/radical/utils/tracer.py
andre-merzky: Ha :)

PR #130, opened 2018-02-06
Feature/json config reader

andre-merzky:
This PR throws away 1k lines of broken configuration code, and replaces it with 200 lines of working config management - or at least that's the idea :-)

This MUST NOT be merged without switching SAGA's configuration management.


andre-merzky:
This PR supports radical-cybertools/saga-python/pull/656.

vivek-bala:
Hey Andre, the changes in the PR look fine by themselves. I am not able to recollect what issue these changes address or how the config module improves usability. I looked at https://github.com/radical-cybertools/saga-python/pull/656/files, but I noticed only a few changes that made use of the config module. Most of them seem to be changes to use the cpi and api modules in saga.

Can you remind me of the issue that this PR resolves? I understand it improves config file mgmt but maybe an example would help me recollect. Thanks!

andre-merzky:
Hey Vivek, 

I can't really remember why I asked you for a review of this one :-P  But thanks a lot for looking into it!

The motivation for this work is three-fold.  The first part is really this:

> This PR throws away 1k lines of broken configuration code, and replaces it with 200 lines of working config management - or at least that's the idea :-)

The code *was* broken, and in fact so much that we could not use it in RP.  The only user of this code is SAGA - so an accompanying effort lives in SAGA which switches to JSON config  files there, too.

The second motivation is SAGA itself: there are many hardcoded system configurations hidden in the SAGA code - those should be moved to config files eventually, similar to what we use as config files in RP.  That is supposed to be based on JSON configs, so that we can share part of the config settings between layers.

Third motivation is RP: the JSON  files there work ok mostly - but they are rather redundant and not very flexible: for example, we have to repeat many parts in order to get a new configuration to work which only differs in one setting (like, `aprun` vs. `orte`).

This PR is supposed to support a simplification of the RP config management.  That though is related to the agent partitioning work (where the config changes are really needed), so there does not yet exist an equivalent PR in RP, yet.

Summary: thanks for looking!  This PR will probably rest for a little longer until the SAGA and RP layers are ready to switch...

andre-merzky:
I'll take you off as a reviewer for now - thanks again.  I'll prompt for a new round of reviews once this all has settled down...

andre-merzky:
See [here](https://github.com/radical-cybertools/saga-python/pull/656) for the related `radical.saga` PR

PR #131, opened 2018-02-12
relocate ID state storage in absence of `$HOME`

andre-merzky:
This recently popped up on titan (see radical-cybertools/radical.pilot/issues/1538)


itomaldonado:
I approve.

PR #141, opened 2018-09-21
handle None messages in watcher

andre-merzky:
This ensures that a `None` type  message on a closing or interrupted endpoint is correctly handled.

### DIFF
--- a/src/radical/utils/process.py
+++ b/src/radical/utils/process.py
@@ -332,12 +332,8 @@ class Process(mp.Process):
                 msg = self._ru_msg_recv(_BUFSIZE)
                 self._ru_log.info('message received: %s' % msg)
 
-                if msg is None:
-                    self._ru_log.warn('received no message, parent closed ep?')
-                    return False
-
-                if msg == '':
-                    self._ru_log.warn('received empty string, parent closed ep!')
+                if msg in [None, '']:
+                    self._ru_log.warn('no message, parent closed ep!')
                     return False
 
                 if msg.strip() == 'STOP':
--- a/src/radical/utils/process.py
+++ b/src/radical/utils/process.py
@@ -331,8 +331,13 @@ class Process(mp.Process):
 
                 msg = self._ru_msg_recv(_BUFSIZE)
                 self._ru_log.info('message received: %s' % msg)
+
+                if msg is None:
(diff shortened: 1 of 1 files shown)


src/radical/utils/process.py
```diff
@@ -331,8 +331,13 @@ def _ru_watch_socket(self):
...
                 msg = self._ru_msg_recv(_BUFSIZE)
                 self._ru_log.info('message received: %s' % msg)
+
+                if msg is None:
+                    self._ru_log.warn('received no message, parent closed ep?')
+                    return False
+
                 if msg == '':
-                    self._ru_log.warn('received empty string, parent closed ep')
+                    self._ru_log.warn('received empty string, parent closed ep!')
```
iparask: What is `ep`?

Do you mind changing `''` to `str()`? I think it is better, because the semantics of an empty string is a bit weird and we do not know how it may change in the future.

Also, why are the `None` and `empty str()` cases different? Wouldn't it be better to have them in the same check with an `or` statement?

iparask:
Also why does the code coverage drop? Shouldn't at least stay the same?

src/radical/utils/process.py
andre-merzky: > Do you mind changing '' to str()? I think it is better, because the semantics of an empty string is a bit weird and we do not know how it may change in the future.

Uh, I'm ok with the others - but I'd rather not, because we don't know what `str()` does in the future either, and it seems no more insane (or sane) to me than `` - we use literal strings all over the place...

> Also, why are the None and empty str() cases different? Wouldn't it be better to have them in the same check with an or statement?

I'll change this -  this was just lazy :-)

andre-merzky:
> Also why does the code coverage drop? Shouldn't at least stay the same?

`\me shrugs`

andre-merzky:
`EP` is Endpoint - the reading or writing end of a socket or pipe.

iparask:
Okay! I see!

vivek-bala:
> Also why does the code coverage drop? Shouldn't at least stay the same?

The codecov probably dropped because 3-4 lines added are not tested (or the test report doesn't capture that)?

iparask:
Okay! Approved by me!

andre-merzky:
Thanks Ioannis, appreciated :-)

PR #142, opened 2018-10-25
Config files for PEP8 and pylint. PEP8 corrections

iparask:
Hello @andre-merzky ,

This PR has the first commits where RU files are passed through PEP8 and the linter. I propose to make the changes incrementally and slowly. That being said there will be a series of PR of this style.


iparask:
Actually this is continuous work. It will not finish very soon. There are cases though that the code still gives me problems. I will ping it when I am done. Feel free though to leave comments whenever you want.

andre-merzky:
We can also merge even if not yet perfect...

PR #147, opened 2018-12-07
Try to fix coverage report on Travis

mturilli:


andre-merzky:
Wouldn't it be better to configure travis in a way that it covers executables instead of changing permissions?   We can't for example change permissions in `bin/` scripts.

PR #151, opened 2019-02-04
Feature/zmq port

andre-merzky:


iparask:
Would it make sense to ask for test? Otherwise the PR seems fine!

andre-merzky:
> Would it make sense to ask for test? Otherwise the PR seems fine!

In principle yes, and I expected that comment.  In practice it is tricky to set up meaningful tests, as that implies the creation of actual communication channels which, as per this PR, needs interaction between different processes - so it does not easily fit into unit tests.  Of course one can do some unit tests on the API level - but they won't tell us much more than that the code is syntactically correct and the pyzmq API is correctly used.

Do you have any feedback on a usable granularity which does not delay this by some weeks?  @mturilli , any suggestion?

iparask:
I would say for now unit tests are more than enough! It would make sure that the zmq API is used correctly and that all parts of the code do what is expected!

andre-merzky:
> that all parts of the code do what is expected

Sorry for not being clear: I think this can only be confirmed with active endpoints, by actually sending from one (API)-end to the other.  Like, you cannot sensibly test a `channel.receive()` method without receiving anything, because the code is basically to `zmq.get`, to unpack the message, and return it, see [here](https://github.com/radical-cybertools/radical.utils/pull/151/files#diff-dc04005ba9f95895ad2810f06b555c45R332).  So what are we testing if there is no message received?  A unit test is not even easily possible since those calls either block when there is no message, or fail if there is no channel, or do nothing.  The unpack of course works on data structures we invent - but do they work on received messages is whats interesting - otherwise we unit-test the unpack module, not our code. 
 Mocking the get and unscramble away leaves an empty method.

I am not trying to be negative here, honestly asking for suggestions.  I think tests on real channels are useful, and event better are protocol level checks - but that is not easy to set up in a unit-test framework, and will take some thinking and time.  Do we want to invest that now?

mturilli:
Does this help? https://stefan.sofa-rockers.org/2012/02/07/designing-and-testing-pyzmq-applications-part-2/

andre-merzky:
Nice link!  They basically say the same though: ZMQ is likely to work, so mock it away; instead, test message handlers and protocols.  But those are not part of this PR - this is really just providing the ZMQ endpoint setup.

mturilli:
Great. Then I would be happy with this PR but I would *concurrently* add the tests to the message handlers and protocols building upon the ideas in that link. Would that work for you?

andre-merzky:
> Great. Then I would be happy with this PR but I would _concurrently_ add the tests to the message handlers and protocols building upon the ideas in that link. Would that work for you?

That's in RP then (nothing in RU uses the channels).  I make this an item for v.2 (there I definitely want to have protocol level testing) and start planning for it.

Thanks both!

PR #156, opened 2019-03-08
Remove black listed from checked

iparask:


andre-merzky:
Thank you!

andre-merzky:
Hmm, travis fails with `invalid syntax` errors outside of our code.  I doubt this has anything to do with the PR - but do you have any idea what's up there?

iparask:
`flake8` and `codecov` does not get installed. Are they in `setup` as requirements? I've seen  it before, and it gets fixed after a couple of hours

PR #159, opened 2019-03-25
fix SAGA issue 305

andre-merzky:
This fix ensures that leading double slashes are collapsed on URL paths, as documented in the SAGA API spec.


tests/unittests/test_url.py
```diff
@@ -38,96 +32,105 @@ def test_url_scheme_issue () :
...
     assert u1.port     == int(9999)
 
+
 # ------------------------------------------------------------------------------
 #
-def test_url_issue_49 () : 
-    """
-    Test url issue #49
-    """
+def test_url_issue_49():
```
mturilli: Should we use descriptive names for the tests instead of the issue number? I think that is common practice? For 305, we would have something like `*_double_slash` or something like that.

tests/unittests/test_url.py
```diff
@@ -38,96 +32,105 @@ def test_url_scheme_issue () :
...
 # ------------------------------------------------------------------------------
 #
-def test_url_properties () :
-    """
-    Test url properties
-    """
+def test_url_issue_rs_305():
+
+    url1 = ru.Url ("advert://localhost/path/to/file")
+    url2 = ru.Url ("advert://localhost//path/to/file")
```
mturilli: Should we test also for:
```
advert://localhost/path//to/file
advert://localhost/path/to//file
```
? More in general, what about `advert://localhost/path/to/file//`?

tests/unittests/test_url.py
andre-merzky: I find `issue` names useful as it's easy to relate to, well, an issue :-)  And names can get long when they are descriptive (`test_url_schema_valid_for_cross_site_copy`).    But a comment with what the test does might make sense (I removed the comments which were only duplicating the name in this file - no value in comments which don't have any information).

tests/unittests/test_url.py
andre-merzky: `os.path.normpath` only struggles with the leading paths, which is the reason why this issue exists:
```
>>> import os
>>> os.path.normpath('//path//to//dir//')
'//path/to/dir'
```
Don't ask me why... :-P

tests/unittests/test_url.py
andre-merzky: BTW: I don't find a method per issue very useful.  What should happen IMO is a set of URL use cases, and a new case being added for this issue.   But that would be out of scope for this PR.

andre-merzky: BTW: I don't find a method per issue very useful. What should happen IMO is a set of URL use cases, and a new case being added for this issue. But that would be out of scope for this PR.

mturilli: I do not mind long names in tests but I am happy to compromise with a descriptive comment :) Agreed on the method per issue but if it works I would not invest further.

andre-merzky:
@mturilli : should be ready too merge?

andre-merzky:
I fixed the test descriptions - ready to merge.

PR #161, opened 2019-04-03
backport expand_env

andre-merzky:
This backports the regex parser and `expand_env` from devel as that's needed to get RP to work on Tiger (don't want to wait for the release) - see radical-cybertools/radical.pilot/pull/1846

### DIFF
--- a/src/radical/utils/misc.py
+++ b/src/radical/utils/misc.py
@@ -1,8 +1,8 @@
 
 import os
+import re
 import sys
 import time
-import regex
 import shlex
 import socket
 import importlib
@@ -10,6 +10,8 @@ import netifaces
 
 import subprocess as sp
 import url as ruu
+
+from .ru_regex import ReString
 
 
 # ------------------------------------------------------------------------------
@@ -158,7 +160,7 @@ def parse_file_staging_directives(directives):
             raise TypeError("file staging directives muct by of type string, "
                             "not %s" % type(directive))
 
-        rs = regex.ReString(directive)
+        rs = ReString(directive)
 
         if  rs // '^(?P<one>.+?)\s*(?P<op><|<<|>|>>)\s*(?P<two>.+)$':
             res = rs.get()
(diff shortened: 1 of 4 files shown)


iparask:
FYI, this is targeted to devel. If the PR is part of a hotfix shouldn't point to master?

src/radical/utils/misc.py
```diff
@@ -512,6 +514,76 @@ def get_env_ns(key, ns, default=None):
     return default
 
 
+# ------------------------------------------------------------------------------
+#
+def expand_env(data, env=None, ignore_missing=True):
```
iparask: Where was this function before? And why it did not work where it was?

src/radical/utils/ru_regex.py
```diff
@@ -4,41 +4,43 @@
...
-    """
+    '''
 
     # --------------------------------------------------------------------------
     #
-    def __init__ (self, result=None) :
+    def __init__(self, result=None):
         """
-        construct with a `re.MatchObject` instance.  This ctor should only be
+        construct with a `regex.MatchObject` instance.  This ctor should only be
```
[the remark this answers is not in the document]
andre-merzky: Not sure I understand - the docstring says what the expected parameter is - a `regex.MatchObject` instance.

src/radical/utils/ru_regex.py
```diff
@@ -4,41 +4,43 @@
...
+        self._glist  = list()
+        self._gdict  = dict()
+        self._result = result
 
 
-        if  result : 
-            if  not isinstance (result, type(re.match("",""))) :  # fuck python 
-                raise TypeError ('Need re.MatchObject on construction, not %s' % type(result))
+        if result:
+            # fuck python
```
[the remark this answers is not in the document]
andre-merzky: I could expand on it ;-)  Because really, this is fucking stupid.  Python claims to be very flexible with types and be great with inspection and all - but when you try to use that outside of some tutorial devel stuff, more often than not that's just broken.  

Case in point:  I want to ensure that the c'tor is getting the right parameter type, because Python's on-the-fly conversion does not work in this context.  But I can't, because the type is not really a type, but a shim, so I have to fake the type check...

src/radical/utils/ru_regex.py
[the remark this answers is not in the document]
iparask: So result is `regex.MatchObject`?

src/radical/utils/misc.py
iparask: Was PR #152 supposed to do that? It is already merged to devel, but when I check devel the `expandvars` function which is the same as the one you introduce here is being removed. When did it get removed and why do we have to reintroduce the same function with different name?

I'm really confused now

andre-merzky: It is exactly that method.  It got *not* removed from devel, it is there: https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/misc.py#L546  This PR is a backport to master because the release is not quite ready yet., but I need this fixed for princeton.tiger support (or at least it makes it easier).

src/radical/utils/ru_regex.py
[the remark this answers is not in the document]
andre-merzky: Yes! :-)

andre-merzky:
> FYI, this is targeted to devel. If the PR is part of a hotfix shouldn't point to master?

Duh - of course that was supposed to go to master, apologies...

src/radical/utils/misc.py
iparask: #152 introduces a function called `expandvars`([here](https://github.com/radical-cybertools/radical.utils/pull/152/files#diff-b6c4e82fd261c1ff617b12c2084dcb9eR625)) which was then renamed to `expand_env` in devel. I am not sure when that happened. Also I know RP master uses the `expandvars` ([here](https://github.com/radical-cybertools/radical.pilot/blob/master/src/radical/pilot/agent/rm/ccm.py#L60)) already, which makes me wonder, because I could not find it in RU master.

I hope this makes clear why I am confused, right?

andre-merzky: That `ru.expandvars` was used in master was wrong - this is exactly what this PR (and the RP one you approved) are fixing.  But yes, I understand the confusion.  This is supposed to fix this :-)

src/radical/utils/ru_regex.py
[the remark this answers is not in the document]
iparask: Still it is not a nice comment!

src/radical/utils/misc.py
andre-merzky: As for the rename: the method now applies to all strings (we want to mostly use it in config files) and has somewhat different syntax and semantics than `os.path.expandvars` - that's why I wanted a name which does not create an association with the `os` method, to avoid confusion.  On the upside: all paths which can be handled by `os.path` can also be handled by our method...

andre-merzky:
Thanks for the review and comments!

PR #163, opened 2019-04-11
extend `expand_env` to dicts and sequences

andre-merzky:
This makes expansion of complete config dicts much simpler, avoids code repetition, and keeps performance overhead localized.


mturilli:
Should we return a warning if an anything other than a dict/list/string is passed for expansion? A silent return may let think that the expansion has been done? I am thinking to a casa in which, for example, I pass a set for expansion.

src/radical/utils/misc.py
```diff
@@ -569,18 +580,44 @@ def expand_env(data, env=None, ignore_missing=True):
...
+
+        for idx, elem in enumerate(data):
+            data[idx] = expand_env(elem, env, ignore_missing)
+        return data
+
+    # all other non-string types are left alone
+    elif not isinstance(data, basestring):
+        return data
+
+    assert(isinstance(data, basestring))
```
mturilli: Why do we need this?

mturilli:
@iparask I see we started to separate the tests in Travis, neat :) In the output of both tests I see the VERY long list of all the linting errors in utils. I know the tests passes because we check the lintig only of the code of the PR: would we be able to avoid printing about all the rest?

src/radical/utils/misc.py
andre-merzky: will remove.

andre-merzky:
> Should we return a warning if an anything other than a dict/list/string is passed for expansion? A silent return may let think that the expansion has been done? I am thinking to a casa in which, for example, I pass a set for expansion.

This is actually on purpose.  Consider this:

```
d = {'file_name' : '$HOME/.radical',
     'file_size' : 1024}
```

I want to expand all strings in this mixed-type dict w/o triggering errors or warnings on whatever cannot be handled (like the `int` above).  This is a common mix we have in our config files which this method tries to support.

andre-merzky:
BTW: a `set` is covered as it is considered a sequence type.

mturilli:
Thanks Andre, yes, I was indeed referring to the case of passing an arbitrary type of collection for expansion and not being told that that specific type of collection is not supported. Basically, I am considering tuples, dicts, sets, frozensets plus those in the collections modules (did I forget any?).

andre-merzky:
Two points: the method will work on anything which has an `__iter__` member, which is what defines a sequence in Python, and documents that other data are left unchanged.  A warning for documented behavior is not useful I would argue?  I would like to avoid getting warnings for the use case this is expected to be used in...

But also: we don't usually do strong type checking on the RU layer.  I don't mind at all to be much more stringent on the user facing APIs (RS, RP, RE etc).

PR #168, opened 2019-07-05
get test config back to work

andre-merzky:
this is used by RS tests


andre-merzky:
@iparask , @AymenFJA , @lee212 : please review - thanks!

PR #172, opened 2019-07-18
Feature/lockfile

andre-merzky:
This is a standalone feature which does not affect other parts of the stack, and thus should be mergeable.  The lockfile class is expected to help simplifying the bootstrapping steps.


mturilli:
I did not merge after approval because most of the tests fail. It is only linting but I am not sure whether they should block the merge or not. Does what we discussed on Tuesday apply only to p3?

PR #175, opened 2019-07-18
Fix/log ns

andre-merzky:
This fixes a bug where logger, profile etc environment was incorrectly evaluated because the configuration was incorrectly pulled from the wrong namespace.


setup.py
```diff
@@ -162,9 +162,10 @@ def read(*rnames):
 # compile gtod
 try:
     compiler = new_compiler(verbose=1)
-    objs = compiler.compile(sources=['src/radical/utils/gtod.c'])
-    exe  = compiler.link_executable(objs, 'bin/radical-utils-gtod')
-except:
+    objs     = compiler.compile(sources=['src/radical/utils/gtod.c'])
+    exe      = compiler.link_executable(objs, 'bin/radical-utils-gtod')
+except Exception as e:
+    print 'gtod compile failed: %s' % repr(e)
```
iparask: This is nice to know!

src/radical/utils/logger.py
```diff
@@ -230,7 +230,7 @@ def __init__(self, name, ns=None, path=None, targets=None, level=None,
             path = ru_def['log_dir']
 
         if not ns:
-            ns = ru_def.get('ns', name)
+            ns = name
```
iparask: I see, you want to make sure that the namespace of the logs are in accordance with the name the logger has. I think it makes sense!

andre-merzky:
> Would it make sense to make travis pass all the checks? I can help if you want

Actually no - I made sure that all tests and lints pass in `python3/devel` - but I am fairly sure that fixing the same problems here will lead to conflicts.  So I'd rather forward this to `python3/devel` once the PR is merged and fix any problems there.  I hope that makes sense...

andre-merzky:
> > Would it make sense to make travis pass all the checks? I can help if you want
> 
> Actually no - I made sure that all tests and lints pass in `python3/devel` - but I am fairly sure that fixing the same problems here will lead to conflicts. So I'd rather forward this to `python3/devel` once the PR is merged and fix any problems there. I hope that makes sense...

PS.: this applies to the linting (which is what fails) - the functional tests should pass on PRs (and they do for this one).

PR #176, opened 2019-07-18
Fix/titan deploy

andre-merzky:
trivial fix to make sure a file is closed before calling chmod on it.


mturilli:
Titan?

andre-merzky:
> Titan?

The name of the branch is historic, the fix is not titan specific :-)

PR #178, opened 2019-08-16
Hotfix/issue rs 731

andre-merzky:
This fixes the problem observed in rs ticket radical-cybertools/radical.saga/issues/731 (amongst others) were logger env is ignored if not specific logger namespace is defined on construction.


andre-merzky:
Thanks! :-)

PR #184, opened 2019-09-17
add daemonization and unit tests

andre-merzky:
Hi @iparask , @mturilli : I am working on process management again as that needs fixing for NGE, and decided to do this in Py3.  This PR is that cleanup + unit test.  The daemonizer works as expected: you can run `python3 tests/unittests/test_damon.py` and it will pass.  However, it will not pass under pytest: pytest will trace the *child* processes which eventually call `sys.exit`, and (a) pytest does not like that, and (b), pytest continues beyond the `sys.exit`, and triggers all kinds  of other errors.

I could not find a way to convince pytest to ignore the subprocesses,  Any idea?


mturilli:
I could not find `tests/unittests/test_damon.py` in either branch.

andre-merzky:
> I could not find `tests/unittests/test_damon.py` in either branch.

Oops, apologies - the file added now.

mturilli:
Would this help? http://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

andre-merzky:
I don't think so.  The problem is that the `sys.exit()` happens in the code which is tested, not in the unit test.  I also can't mock `sys.exit()` away, as that would screw with the semantic of the code (that process needs to exit there).

mturilli:
Got it. Roaming for a solution I found something like:
```
def test_exit(mymodule):
    with pytest.raises(SystemExit) as pytest_wrapped_e:
            mymodule.will_exit_somewhere_down_the_stack()
    assert pytest_wrapped_e.type == SystemExit
    assert pytest_wrapped_e.value.code == 42
```
Do the use of a named exception and the wrapping into an object go in the right direction? (Credit: https://medium.com/python-pandemonium/testing-sys-exit-with-pytest-10c6e5f7726f)

andre-merzky:
@iparask , @mturilli : I'll merge this now, too many conflicts piling up on depending branches...

PR #185, opened 2019-09-18
support callbacks for heartbeat management

andre-merzky:
This PR adds heartbeat callbacks, to simplify lifetime management in the RCT stack.  Specifically, an RP component or session can use those callbacks to regularly send heartbeats on the HB channel.


andre-merzky:
@lee212 : I need to merge this, too many conflicts piling up in merges which depend on this...

PR #187, opened 2019-10-09
fix int and float parsing during env expansion

andre-merzky:
this fixes https://github.com/radical-cybertools/radical.saga/issues/739


iparask:
@andre-merzky, this is a python2 branch. I cannot install it to a python3 env.

andre-merzky:
Ah bugger - let me sync (it should merge cleanly)

andre-merzky:
This is py3 based now.

iparask:
I think the target branch should also be changed! :)

andre-merzky:
> I think the target branch should also be changed! :)

It is already :-)  Probably crossed notifications...

iparask:
This works

andre-merzky:
Thanks!

PR #191, opened 2019-11-06
Feature/proc mgmt

andre-merzky:
This PR supports the switch from watcher + signal based process management to a heartbeat based one.  Unfortunately it touches several different parts of RU.  The most important changes are

  - remove `ru.Process` and `ru.Thread` classes
  - add startup synchronization for heartbeats
  - add callbacks to `ru.Heartbeat` instances (on events and termination)
  - have ZMQ bridges host a heartbeat instance
  - sandbox `ru.Daemon` instances
  - isolate ZMQ bridge and channel configuration

There are a number of related minor changes though, which mostly center around configuration, logging and debugging:

  - add logging capabilities to daemonized processes and bridges
  - simplify debug helper (reduce `ru.Process` support), improve debug of thread locks
  - fixes to the configuration system, specifically to config merges and type preservation / conversion
  - API consistency, specifically wrt. namespaces on logging, profiling, debugging and configuration

I'd be happy to interactively help with the review process, just let me know.


src/radical/utils/atfork/stdlib_fixer.py
```diff
@@ -54,7 +54,7 @@ class Error(Exception):
 
 def fix_logging_module():
 
-    if 'RADICAL_UTILS_NOATFORK' in os.environ:
+    if 'RADICAL_UTILS_PATCHATFORK' not in os.environ:
         return
```
mturilli: Is this intended?

src/radical/utils/config.py
```diff
@@ -119,116 +122,168 @@
...
 
-import glob
 import os
+import glob
+import munch
 
 from .misc       import find_module, is_string
 from .misc       import expand_env as ru_expand_env
-from .read_json  import read_json
+from .read_json  import read_json, write_json
```
mturilli: Should we change the name seeing that we do both reading and writing?

src/radical/utils/config.py
```diff
@@ -6,6 +6,8 @@
 
 # ------------------------------------------------------------------------------
 #
+# FIXME: documentation is out of sync with implementation
```
mturilli: Probably this needs to be fixed before merging?

src/radical/utils/config.py
```diff
@@ -119,116 +122,168 @@
...
+        For example, if the following is used in a system python installation:
+
+            ru.Config('radical.pilot', category='session', name='minimal')
+
+        it would attempt to load (depending on system details):
+
+            /usr/lib/python3/site-packages/radical/pilot/\
+                                                   config/session_mininmal.json
+
+        NOTE: keys MUST NOT begin with an underscore, and MUST NOT contain dots.
```
mturilli: Do we have a sanity check with informative error/warning for these two MUST NOT?

src/radical/utils/config.py
```diff
@@ -237,84 +292,147 @@ def __init__(self, module, path=None, name=None, cfg=None,
...
                 for usr_fname in glob.glob(usr_fspec):
+
                     base = usr_fname[prefix_len:-postfix_len]
                     ucfg = read_json(usr_fname)
                     usr_cfg[base] = ucfg
+                    nfiles += 1
+
+        # if we did not find *any* file, and the original `name` was None,
+        # then try to load config files w/o name
+        # eg, if there is no `registry_default.json`, then try to load
```
mturilli: eg, -> e.g.,

src/radical/utils/config.py
```diff
@@ -237,84 +292,147 @@ def __init__(self, module, path=None, name=None, cfg=None,
...
+
+        # merge sys, usr and app cfg before expansion
+        cfg_dict = dict()
+        cfg_dict = dict_merge(cfg_dict, sys_cfg, policy='overwrite')
+        cfg_dict = dict_merge(cfg_dict, usr_cfg, policy='overwrite')
+        cfg_dict = dict_merge(cfg_dict, app_cfg, policy='overwrite')
+
+        if cfg_dict:
+
+            # ------------------------------------------------------------------
```
mturilli: Why this function is declared after a guard?

src/radical/utils/atfork/stdlib_fixer.py
andre-merzky: Yes - `atfork` was part of the old proc mgmt, so instead of opt-out it is now opt-in.  I did not remove completely since we might need it for profile flushing, I would like to watch how that behaves for a while.

src/radical/utils/config.py
andre-merzky: Not sure I understand - change the name how?

src/radical/utils/config.py
andre-merzky: ack

src/radical/utils/config.py
andre-merzky: No.  That needs to be clearly documented, but we can also asserts those conditions.

src/radical/utils/config.py
andre-merzky: because it's only needed in that case.  Moving the helper function elsewhere would work and not change anything semantically, but I wanted to keep it close to where it is being used (the line below).

src/radical/utils/config.py
mturilli: `from .read_json  import read_json, write_json` -> `from .?json_utils? import read_json, write_json` or something like that?

src/radical/utils/config.py
mturilli: Yes, I would put that in a guard/asset/whatever caught with a dedicated exception. Just thinking about avoiding debugging when the user unavoidably will use forbidden characters.

src/radical/utils/config.py
andre-merzky: Ah, you mean the filename - sorry, missed that!  Yeah, I guess that could be done.  Can't be `json` though, that conflicts with the Python module of the same name.  I'll think of something...

src/radical/utils/__init__.py
```diff
@@ -13,18 +13,16 @@
...
 from .threads        import main_thread, this_thread, get_thread_name, gettid
 from .threads        import set_cancellation_handler, unset_cancellation_handler
 from .threads        import raise_in_thread, ThreadExit, SignalRaised
 from .futures        import Future
 from .futures        import NEW, RUNNING, DONE, FAILED, CANCELED
 from .url            import Url
-from .dict_mixin     import DictMixin, dict_merge, dict_stringexpand
-from .dict_mixin     import PRESERVE, OVERWRITE
+from .dict_mixin     import DictMixin, dict_merge, dict_stringexpand, dict_diff
+from .dict_mixin     import PRESERVE, OVERWRITE, iter_diff
```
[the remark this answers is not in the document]
andre-merzky: Apologies, I am not sure what you mean with 'merge imports'?  They can't be all on one line, that line is then too long?

src/radical/utils/config.py
andre-merzky: it is `json_io.py` now.

src/radical/utils/config.py
```diff
@@ -328,27 +446,30 @@ def query(self, key, default=None):
         this method behaves like:
 
             config['some']['path']['to'].get('key', default='foo')
-
         '''
+        # TODO: overload `get()`?
```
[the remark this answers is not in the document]
andre-merzky: `get(key, default)` would give a convenient way to query with a default value.

src/radical/utils/debug.py
```diff
@@ -129,18 +46,37 @@ def print_stacktraces(signum=None, sigframe=None):
...
     # the hope that this will stagger the prints.
     pid = int(os.getpid())
     time.sleep((pid % 100) / 1000)
 
-    out  = '===============================================================\n'
+    out  = '=========================================================\n'
     out += 'RADICAL Utils -- Debug Helper -- Stacktraces\n'
     out += os.popen("%s | grep ' %s ' | grep -v grep" % (_ps_cmd, pid)).read()
 
+    if _debug_helper:
```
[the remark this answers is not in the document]
andre-merzky: It's defined elsewhere in the same file.

src/radical/utils/dict_mixin.py
```diff
@@ -132,8 +126,8 @@ def dict_merge(a, b, policy=None, wildcards=False, logger=None, _path=None):
 
     '''
 
-    if  a    is None: return
-    if  b    is None: return
+    if  a    is None: return a
+    if  b    is None: return a
```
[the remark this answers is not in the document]
andre-merzky: Yes - the call always returns a with be being merged into it.  If a is None, then None is returned (== a).  If b is None, then a is returned unchanged.

src/radical/utils/ids.py
```diff
@@ -139,16 +139,15 @@ def generate_id(prefix, mode=ID_SIMPLE, namespace=None):
...
+    local to that namespace, . For example, you can create a session and use the
+    session ID as a namespace for all the IDs of the objects of that execution.
 
     Example::
 
         sid  = generate_id('re.session', ID_PRIVATE)
-        uid1 = generate_id('task.%(item_counter)04d', ID_CUSTOM, namespace=sid)
-        uid2 = generate_id('task.%(item_counter)04d', ID_CUSTOM, namespace=sid)
+        uid1 = generate_id('task.%(item_counter)04d', ID_CUSTOM, ns=sid)
+        uid2 = generate_id('task.%(item_counter)04d', ID_CUSTOM, ns=sid)
```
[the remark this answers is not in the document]
andre-merzky: Not sure I understand - `sid` is an arbitrary string, and can be set to `foo`.

src/radical/utils/logger.py
```diff
@@ -77,6 +77,7 @@ def release_all(self):
                 logger = logger.parent
 
     def close_all(self):
+        # FIXME: add `atexit` handler
```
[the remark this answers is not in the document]
andre-merzky: To flush logfiles before closing.  This might not be needed, but we should observe the flushing behavior for a while to ensure the new process handling did not screw with logfile flushing.  If need be, the flushing needs to be added as `atexit` handler (which would call `close()` explicitly I assume).

src/radical/utils/plugins/unittests_2/plugin_unittests_default_2.py
```diff
@@ -12,36 +12,29 @@
     'type'        : 'unittests_2',
     'name'        : 'default_2',
     'version'     : '0.1',
-    'description' : 'this is an empty test which basically does nothing.'
+    'description' : 'this is a test which basically does nothing.'
```
mturilli: Should we briefly explain why we have it then?

src/radical/utils/__init__.py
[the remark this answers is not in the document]
andre-merzky: @iparask : ping

src/radical/utils/zmq/bridge.py
```diff
@@ -1,9 +1,73 @@
...
+import msgpack
+
+import threading as mt
 
 from ..logger    import Logger
 from ..profile   import Profiler
-from ..heartbeat import Heartbeat
+
+
+_MAX_RETRY = 3  # max number of ZMQ snd/rcv retries on interrupts
```
mturilli: Wondering whether this should be 'elevated' to a configurable parameter

src/radical/utils/zmq/bridge.py
```diff
@@ -1,9 +1,73 @@
...
+
+
+# --------------------------------------------------------------------------
+#
+# zmq will (rightly) barf at interrupted system calls.  We are able to rerun
+# those calls.
+#
+# FIXME: how does that behave wrt. timeouts?  We probably should include
+#        an explicit timeout parameter.
+#
```
mturilli: As part of 'bridge conf' parameters retry and timeout?

src/radical/utils/zmq/bridge.py
```diff
@@ -19,53 +83,56 @@ class Bridge(object):
...
-        this *must* be called by deriving classes
-        '''
+    def start(self):
 
-        self._hb.beat()
+        # the bridge runs in a thread.  It is the bridge's owner process'
+        # responsibility to ensure the thread is seeing suffient time to perform
+        # as needed.  Given Python's thread performance (or lack thereof), this
+        # basically means that the user of this class should create a separate
+        # process instance to host the bridge thread.
```
mturilli: Why we do not run it directly in a dedicated process?

tests/unittests/test_misc.py
```diff
@@ -108,23 +108,27 @@ def test_get_env_ns():
 #
 def test_expand_env():
 
-    noenv = {'FOO' : 'foo'}
+    noenv = {'BIZ' : 'biz'}
     env   = {'BAR' : 'bar'}
```
mturilli: Could we have a more descriptive filename that 'test_misc.py'?If not, then maybe the semantic of what is tested is not clear enough?

src/radical/utils/config.py
andre-merzky: done

src/radical/utils/config.py
andre-merzky: I corrected the doc string - the keys are not really forbidden, but interpreted differently (underscore is used for internal bookkeeping, dots are path separators).

src/radical/utils/config.py
andre-merzky: fixed

src/radical/utils/config.py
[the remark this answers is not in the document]
andre-merzky: This is done by now, I removed the comment.

src/radical/utils/config.py
```diff
@@ -328,27 +446,30 @@ def query(self, key, default=None):
...
 
         if is_string(key): elems = key.split('.')
         else             : elems = key
 
         if not elems:
             raise ValueError('empty key on query')
 
-        pos  = self._cfg
+        pos  = self
         path = list()
```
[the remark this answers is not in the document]
andre-merzky: It is used for the error message.

src/radical/utils/heartbeat.py
```diff
@@ -34,48 +50,112 @@ def __init__(self, uid, timeout, interval=1, cb=None, term=None, log=None):
...
-                    os.kill(self._pid, signal.SIGTERM)
+                                       self._uid, uid, now, last, self._timeout)
+
+                    ret = False
+                    if  self._term_cb:
+                        self._log.warn('hb term cb for %s: %s', uid, self._term_cb)
+                        try:
+                            ret = self._term_cb(uid)
+                        except:
+                            self._log.exception('=== oops')
```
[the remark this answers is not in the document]
andre-merzky: Thanks - I removed some log messages which were left over from debug sessions.

src/radical/utils/heartbeat.py
```diff
@@ -34,48 +50,112 @@ def __init__(self, uid, timeout, interval=1, cb=None, term=None, log=None):
...
+                    if ret is False:
+                        # could not recover: abandon mothership
+                        self._log.warn('hb failure for %s - fatal (%d)', uid,
+                                self._pid)
+                        os.kill(self._pid, signal.SIGTERM)
+                        time.sleep(1)
+                        os.kill(self._pid, signal.SIGKILL)
+
+                    else:
+                        # recovered - the failed UID wath replaced with the one
```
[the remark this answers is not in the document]
andre-merzky: Thanks - fixed ("was" :-) )

src/radical/utils/ids.py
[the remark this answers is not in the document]
andre-merzky: @iparask : ping

src/radical/utils/json_io.py
```diff
@@ -39,19 +41,24 @@ def read_json_str(filename):
...
 #
-def write_json(data, filename):
+def write_json(data, fname):
     '''
     thin wrapper around python's json write, for consistency of interface
 
     '''
 
+    if isinstance(fname, dict) and isinstance(data, str):
+        # be generous
```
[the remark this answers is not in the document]
andre-merzky: comment added.

src/radical/utils/misc.py
```diff
@@ -318,8 +318,9 @@ def as_list(data):
     return non-list data into a list.
     '''
 
-    if is_list(data): return data
-    else            : return [data]
+    if   data is None : return []
+    elif is_list(data): return data
+    else              : return [data]
```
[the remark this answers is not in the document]
andre-merzky: A list does not need conversion to a list :-)

src/radical/utils/misc.py
```diff
@@ -359,13 +360,18 @@ def is_string(data):
...
+
+    elif isinstance(data, list):
+        return [as_string(e) for e in data]
+
+    elif isinstance(data, bytes):
+        return bytes.decode(data, 'utf-8')
+
     else:
-        raise TypeError('cannot convert %s to string' % type(data))
+        return data
```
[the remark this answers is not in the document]
andre-merzky: This method only converts bytearreays to strings, but leaves all other data types alone (but iterates through dicts and lists).  So leaving bools and `None` etc. unchanged is the right behavior.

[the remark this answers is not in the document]
andre-merzky: Comment added.

src/radical/utils/misc.py
```diff
@@ -384,16 +390,21 @@ def is_bytes(data):
...
+
+    elif isinstance(data, list):
+        return [as_bytes(e) for e in data]
+
+    elif isinstance(data, str):
+        return str.encode(data, 'utf-8')
+
     else:
-        raise TypeError('cannot convert %s to bytes' % type(data))
+        return data
```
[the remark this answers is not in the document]
andre-merzky: ack.

src/radical/utils/plugins/unittests_2/plugin_unittests_default_2.py
andre-merzky: It says so:  it is a test item that has no useful semantics.

src/radical/utils/zmq/bridge.py
andre-merzky: I  don't think this would be meaningful to the user, and it's difficult to document what a good value is.

src/radical/utils/zmq/bridge.py
andre-merzky: This question applies to calls like `get_nowait(timeout=seconds)`.  The call will be handled semantically correct even if interrupted and recovered, but I did not check how ZMQ handles the timeout parameter then, like, if the timeout is applied again on recover, or if the total timeout is respected (timeout of initial attempt and timeout for recover attempt get added).  I assume the latter - but it's not important enough to actually test and fix (at the moment), so I just left a note.

src/radical/utils/zmq/bridge.py
andre-merzky: Well, we do - but the process should be spawned by the user of this class, which is what this comment tries to convey.  The process management lives on top of the communication channel management (otherwise we could not use ZMQ to sync between processes (heartbeats).

tests/unittests/test_misc.py
andre-merzky: hehe - this *is* the descriptive name, as it tests functionality provided by `src/radical/utils/misc.py` :-)

src/radical/utils/zmq/pubsub.py
```diff
@@ -97,95 +43,75 @@ def uid(self):
...
 
     @property
     def addr_in(self):
-        return self._addr_in
+        return self.addr_pub
 
     @property
     def addr_out(self):
-        return self._addr_out
+        return self.addr_sub
```
mturilli: Are you trying to keep compatibility with the previous use of *_in/*_out? I see that in the following code you transition from *_in to *_pub. If so, should we just clean the whole interface out and use only *_pub/*_sub?

src/radical/utils/zmq/pubsub.py
```diff
@@ -280,38 +209,129 @@ def channel(self):
...
+
+                topic, msg = Subscriber._get_nowait(socket, lock, 500, channel)
+                if topic and channel == 'state_pubsub':
+                    log.debug('get %s [%s]', topic, len(msg))
+
+                if topic:
+                    t = as_string(topic)
+                    for m in as_list(msg):
+                        m = as_string(m)
+                        for cb, l in callbacks:
```
mturilli: what is `l` here? log? Would avoid too many single letter variables.

src/radical/utils/zmq/pubsub.py
```diff
@@ -280,38 +209,129 @@ def channel(self):
...
+                          # log.debug('cb  %s [%s] [%s]', cb, len(msg), l)
+                            if l:
+                                with l:
```
mturilli: I am not sure I fully understand the use of `with` in this context.

src/radical/utils/zmq/pubsub.py
```diff
@@ -280,38 +209,129 @@ def channel(self):
...
+
+    # --------------------------------------------------------------------------
+    #
+    def __init__(self, channel, url, topic=None, cb=None, log=None):
+        '''
+        If a `topic` is given, the channel will subscribe to that topic
+        immediately.
+
+        When a callback `cb` is specified, then the Subscriber c'tor will spawn
+        a separate thread which continues to listen on the channel, and the cb
```
mturilli: Is this still true? We have a single thread for all the cb right? BTW, c'tor == communicator?

src/radical/utils/zmq/pubsub.py
```diff
@@ -280,38 +209,129 @@ def channel(self):
...
+        self._lock     = mt.Lock()
         self._ctx      = zmq.Context()  # rely on GC for destruction
-        self._q        = self._ctx.socket(zmq.SUB)
-        self._q.linger = _LINGER_TIMEOUT
-        self._q.hwm    = _HIGH_WATER_MARK
-        self._q.connect(self._url)
+
+        if url not in Subscriber._callbacks:
+
+            s        = self._ctx.socket(zmq.SUB)
```
mturilli: Can we use sckt instead of s? We also have subscribe(r) in the same context.

src/radical/utils/zmq/pubsub.py
```diff
@@ -331,22 +351,63 @@ def channel(self):
...
     #
-    def subscribe(self, topic):
+    def _start_listener(self):
 
-        topic = topic.replace(' ', '_')
+        # only start if needed
+        if Subscriber._callbacks[self._url]['thread']:
+            return
+
+        t = mt.Thread(target=Subscriber._listener, args=[self._url, self._log])
```
mturilli: thrd?

src/radical/utils/zmq/pubsub.py
```diff
@@ -331,22 +351,63 @@ def channel(self):
...
+    #
+    def subscribe(self, topic, cb=None, lock=None):
+
+        # if we need to serve callbacks, then open a thread to watch the socket
+        # and register the callbacks.  If a thread is already runnning on that
+        # channel, just register the callback.
+        #
+        # Note that once a thread is watching a socket, we cannot allow to use
+        # `get()` and `get_nowait()` anymore, as those will interfere with the
+        # rhread consuming the messages,
```
mturilli: rhread -> thread?

src/radical/utils/zmq/pubsub.py
```diff
@@ -331,22 +351,63 @@ def channel(self):
...
+        # The given callback (if any) is used to shield concurrent cb
+        # invokations.
+
+        if cb:
+
+            self._interactive = False
+            self._start_listener()
+            Subscriber._callbacks[self._url]['callbacks'].append([cb, lock])
+
+        s = Subscriber._callbacks[self._url]['socket']
```
mturilli: sbscrbr?

src/radical/utils/zmq/pubsub.py
```diff
@@ -355,26 +416,17 @@ def get(self):
...
     #
-    def get_nowait(self, timeout=None):  # timeout in ms
-
-        if _uninterruptible(self._q.poll, flags=zmq.POLLIN, timeout=timeout):
+    def get_nowait(self, timeout=None):
 
-            data  = _uninterruptible(self._q.recv_multipart, flags=zmq.NOBLOCK)
-            topic = None
+        if not self._interactive:
+            raise RuntimeError('get_nowait() on non-interactive EP')
```
mturilli: EP == endpoint? In a raise I would use a message as explicit as possible.

src/radical/utils/zmq/queue.py
```diff
@@ -161,98 +106,76 @@ def uid(self):
         return self._uid
 
     @property
-    def channel(self):
-        return self._channel
+    def type_in(self):
+        return 'put'
+
```
mturilli: See my comment in subpub about abandoning *_in/*_out. Should we have a base class inherited by pubsub/queue or whatever other messaging 'pattern'?

src/radical/utils/zmq/queue.py
```diff
@@ -265,59 +188,60 @@ def _bridge_work(self):
         try:
 
             buf = list()
-
-            i = -1
+            i   = -1
```
mturilli: What is `i`?

src/radical/utils/zmq/bridge.py
mturilli: I am not necessarily thinking to expose it to the user but I am thinking more to the idea of using RP as a framework. Exactly because we (as in devops) we may want to tweak this value, depending on the contingency of workload/resource coupling, I would want to have all these parameters in a single place. When they are scattered across all the components/files, basically you are the only one to have the change (until your memory fails you :)) to know that all these tweaks are even possible.

src/radical/utils/plugins/unittests_2/plugin_unittests_default_2.py
mturilli: Should we keep it then?

src/radical/utils/heartbeat.py
```diff
@@ -1,31 +1,47 @@
...
+        gets a single argument, the uid of the failing heartbeat sender.  If
+        that term callback returns `True`, the kill is avoided though: timers
+        are reset and everything continues like before.  This should be used to
+        recover from failing components.
         '''
 
+        # we should not need to lock timestamps, in the current CPython
+        # implementation, dict access is assumed to be atomic.  But heartbeats
+        # should not be in the performance critical path, and should not have
+        # threads competing with the (provate) dict lock, so we accept the
```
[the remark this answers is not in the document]
andre-merzky: thanks, fixed.

src/radical/utils/plugins/unittests_2/plugin_unittests_default_2.py
andre-merzky: Well, it is a test, so I would think we keep it?

I think I am not clear enough, let me try again.  RU provides a plugin manager.  In order to test it, we need a plugin.  That test plugin does not need to do anything, we just need it so we can test if the plugin manager can load it.  This file (`plugins/unittests_2/plugin_unittests_default_2.py`) is one of the test plugins.  It's description says as much: this is a test and does not do anything, really.  We need to keep it to test the manager.

Did that make more sense?

src/radical/utils/zmq/bridge.py
andre-merzky: Understood.  But I don't think this should be tweaked ever, unless it breaks.  It's only has a defined name because I dislike magic numbers in any code - this way it will pop up quicker and clearer during debugging.  If this is too convoluted, I can move the literal `3` into the code.

Why 3: 1 interrupt on a specific call should be rare, but can happen.  2 points to a very busy system, and is a sign of trouble -- but it can happen just by chance.  Like in order of one per million I would guess, at least for the type of messages we use ZMQ for.  So we pick 3 to be on the save side.  This never triggered in the past AFAIK.

src/radical/utils/zmq/pubsub.py
andre-merzky: If you have a Bridge and know the bridge type is PubSub, then you know the channels support `pub` and `sub` on in/out ports.  But sometimes we instantiate bridges on behalf of applications, and don't keep track of the bridge type (at least not explicitly), and don't expect the user to do so either, and then allow the generic address query for in/out ports, no matter if it is a PubSub bridge (pub / sub) or a Queue (put / get).

I don't mind removing them if you think that is too much cognitive overload - our own code does not really use add_in / addr_out, FWIW.

src/radical/utils/zmq/pubsub.py
andre-merzky: `lock` - the var name is already taken in that scope, but I renamed now to `_lock`.

src/radical/utils/zmq/pubsub.py
andre-merzky: `with lock` acquires the thread lock, and releases it when the scope is left.  Lock release by code indentation - that's  Python :-P

src/radical/utils/zmq/pubsub.py
andre-merzky: What is described in this place is true - the thread is spawned in line 354.  Callbacks need separate threads, otherwise they cannot be truly asynchronous.  Having said that, if the application registers additional callbacks on the same channel later on, they are all served from the same listener thread.

src/radical/utils/plugins/unittests_2/plugin_unittests_default_2.py
mturilli: Yes, now I understand it. Maybe I did not see it because it was hidden by the code review (there were too many files to do the review by expanding all the files). If what you just wrote is not written in the docstring of the test, could you add it?

src/radical/utils/zmq/pubsub.py
andre-merzky: Well,  `t` is used in exactly 3 lines, all together, and has nothing in between.  The type is `Tread`, explicit in the first line.  I don't think it's possible to have any confusion here.

I am not worried about this specific instance - but that paradigm is used in many places in the code.  Single letter vars should only be used in very localized and clear contexts though -- but then I find them useful and concise

```
for f in filenames:
    assert(os.path.isfile(f))

for x in range(spacing_x):
    for y in range(spacing_y):
        dist = dist(zero_x, x, zero_y, y)

t = Thread(...)
t.start()

s = Socket()
setsockopt(s, ...)

```

src/radical/utils/zmq/pubsub.py
andre-merzky: thanks, fixed.

src/radical/utils/zmq/pubsub.py
andre-merzky: `sock` - fixed.

src/radical/utils/zmq/pubsub.py
andre-merzky: see below

src/radical/utils/zmq/bridge.py
mturilli: OK, thanks for explaining this. Seeing that you wrote it already, should we add this to the code as comment? For the rest, this is fine as it is for me now.

src/radical/utils/zmq/pubsub.py
andre-merzky: thanks, fixed.

src/radical/utils/zmq/queue.py
andre-merzky: `put / get` for the queue - but same reply from pubsub applies.

src/radical/utils/zmq/queue.py
andre-merzky: It's not used anymore, left over from debugging - I removed it.

src/radical/utils/zmq/pubsub.py
mturilli: Thanks, maybe add this as a comment? I like the option now that I understand it and I think it is a nice 'service' for the potential user. I would therefore document this and leave it in. What do you think?

src/radical/utils/zmq/pubsub.py
mturilli: Thanks!

src/radical/utils/zmq/pubsub.py
mturilli: Thanks!

src/radical/utils/zmq/pubsub.py
mturilli: Yes, I saw the comment in the following method so this is fine now. Thanks.

src/radical/utils/plugins/unittests_2/plugin_unittests_default_2.py
andre-merzky: The class docstring says:
```
This class implements the (empty) default unittest plugin for radical.utils
```
and it is in a file tree which marks it as unit test, too - sounds ok for you?

mturilli: Sorry, I do not know how to propose a push as @iparask did. Here a suggestion for a C&P to be added to that docstring if you like it: "The default unittest plugin does nothing and it is used, for example, to test whether the plugin manager can load a plugin."

src/radical/utils/ids.py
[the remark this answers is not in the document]
iparask:
```
This will generate the following ids::
        re.session.rivendell.vivek.017548.0001
        task.0000
        task.0001
        where the `task.*` IDs are unique for the used sid namespace.
```
An `sid` is passed, but the example outputs do not include it. Can we include an example that shows how the output changes when `sid` is set.

src/radical/utils/logger.py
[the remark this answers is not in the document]
iparask: May I propose to say `add ataxia handler to flush logfiles before closing`?

src/radical/utils/__init__.py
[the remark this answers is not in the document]
iparask: Oh, that is why you broke it. Okay then.

src/radical/utils/debug.py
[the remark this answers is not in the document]
iparask: Does it make sense, since it is a global variable, to move its definition at the top of the file?

src/radical/utils/dict_mixin.py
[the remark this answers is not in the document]
iparask: Okay. Thank you!

src/radical/utils/ids.py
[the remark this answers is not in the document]
andre-merzky: `sid` is passed as namespace - the returned UID's don't contain that sid, but they are unique in that namespace.  That is how we have unique task IDs *per session*.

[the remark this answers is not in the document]
andre-merzky: BTW: this is also documented just above:
```
    'ns' argument can be specified to a value such that unique IDs are created
    local to that namespace. For example, you can create a session and use the
    session ID as a namespace for all the IDs of the objects of that execution.
```

src/radical/utils/logger.py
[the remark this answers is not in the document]
andre-merzky: I removed the comment as this is currently not needed.

src/radical/utils/plugins/unittests_2/plugin_unittests_default_2.py
andre-merzky: With all due respect, this is a bit of a bike shed discussion.  While I see how this is not really obvious out of context as in this PR, it is in a *plugins/unittest...* directory, in a  file named  *plugin_unittest...*, typed as *plugin type: unittest_2*, and described as being a *test* - I don't think it needs more documentation saying 'it is for testing plugins'.  I am hesitant to just do it because there are several of those test plugins and obviously many other unit tests, and I am too lazy to go though all of them adding `"test_abc" is for testing "abc"`...

src/radical/utils/zmq/bridge.py
andre-merzky: done

src/radical/utils/zmq/pubsub.py
andre-merzky: done

src/radical/utils/plugins/unittests_2/plugin_unittests_default_2.py
mturilli: Sure, I trust your judgement on this.

mturilli:
Thank you for addressing all the comments. I think we have two outstanding general comments: zmq-related tests and Travis failing all but 1 tests. @andre-merzky Would you agree we should address both these issues before merging?

andre-merzky:
Travis needs to pass for sure, I'll take care of that.  ZMQ tests: I can't find that comment anymore, can you point me to it?  But either way, with tests I would appreciate some help: that would involve presumably running a bridge, a producer and consumer, and send some messages I guess.  The core of the PR though is lifetime management via heartbeats, which basically will cause the bridge to die once the parent process dies - but that mechanism actually lives on the RP level, not in RU.  Tests for that heartbeat mechanism itself though are included [here](https://github.com/radical-cybertools/radical.utils/blob/devel/tests/unittests/test_heartbeat.py) (this PR touches those tests, too).  Do you consider that part sufficient?   What other tests would you consider suitable / necessary?

src/radical/utils/zmq/pubsub.py
```diff
@@ -195,44 +123,42 @@ def _bridge_work(self):
...
 
         try:
 
-            while True:
+            while not self._term.is_set():
 
                 # timeout in ms
-                _socks = dict(_uninterruptible(self._poll.poll, timeout=1000))
-                active = False
+                _socks = dict(no_intr(self._poll.poll, timeout=500))
```
iparask: Does the heartbeat repeat every half a second?

src/radical/utils/zmq/pubsub.py
```diff
@@ -280,38 +211,129 @@ def channel(self):
...
-        data  = msgpack.packb(msg)
+        assert(isinstance(topic, str )), 'invalid topic type'
+        assert(isinstance(msg,   dict)), 'invalid message type'
 
+        self._log.debug('-> %s: %s', self.channel, msg)
         log_bulk(self._log, msg, '-> %s' % self.channel)
 
-        msg = [topic, data]
-        _uninterruptible(self._q.send_multipart, msg)
+        btopic = as_bytes(topic.replace(' ', '_'))
```
iparask: Is this to avoid escape characters while the message is sent? If yes, when the message is received do you change `'_'` with `' '`? Would that create an issue to a message that includes from the beginning an `'_'`?

src/radical/utils/zmq/pubsub.py
```diff
@@ -331,22 +353,62 @@ def channel(self):
...
+        with self._lock:
+            no_intr(sock.setsockopt, zmq.SUBSCRIBE, as_bytes(topic))
 
 
     # --------------------------------------------------------------------------
     #
     def get(self):
 
+        if not self._interactive:
+            raise RuntimeError('invalid get(): callbacks are registered')
```
iparask: I'm a bit confused with this method. When a sub is subscribed to a topic, it get out of interactive mode and this methods checks it. Is there a way to use `get` or something similar while the subscriber is in interactive mode?

andre-merzky:
> This is a very big and difficult PR. I'll have more comments tomorrow.

Ha, wait 'til you see the RP one :-P  Seriously though: I know, apologies - but we always knew that touching termination will be a big one *sigh*.  I can only repeat the offer for an interactive review / walk-through, if you think that would help.

src/radical/utils/debug.py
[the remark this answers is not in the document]
andre-merzky: Alas, the `DebugHelper` class needs to be defined first before it can be instantiated, so it cannot be at the top.

src/radical/utils/profile.py
```diff
@@ -46,6 +48,51 @@
...
+
+def _sync_ntp():
+
+    # read from disk cache
+    try:
+        with open('%s/ntp.cache' % get_radical_base('utils'), 'r') as fin:
+            data  = as_string(fin.read()).split()
+            t_sys = float(data[0])
+            t_ntp = float(data[1])
+            t_now = time.time()
```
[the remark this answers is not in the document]
andre-merzky: I guess - but that file is neither locked nor a scarce resource, so it makes not difference at all.  I don't see an advantage - do you?

src/radical/utils/zmq/bridge.py
```diff
@@ -1,9 +1,74 @@
...
+
+    if isinstance(bulk[0], dict) and 'arg' in bulk[0]:
+        bulk = [e['arg'] for e in bulk]
+
+    if isinstance(bulk[0], dict) and 'uid' in bulk[0]:
+        for e in bulk:
+            log.debug("%s: %s [%s]", token, e['uid'], e.get('state'))
+    else:
+        for e in bulk:
+            log.debug("%s: ?", str(token))
```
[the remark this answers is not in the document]
andre-merzky: This is a rather expansive debug log helper - if the message contains dicts with a state and uid (things like an RP task or pilot etc), it logs the type, uid and state, otherwise it just logs an `?` to mark it as an opaque message.

src/radical/utils/zmq/bridge.py
```diff
@@ -19,53 +84,56 @@ class Bridge(object):
...
-        '''
+    def start(self):
 
-        self._hb.beat()
+        # the bridge runs in a thread.  It is the bridge's owner process'
+        # responsibility to ensure the thread is seeing suffient time to perform
+        # as needed.  Given Python's thread performance (or lack thereof), this
+        # basically means that the user of this class should create a separate
+        # process instance to host the bridge thread.
+        self._term          = mt.Event()
```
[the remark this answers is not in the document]
andre-merzky: Yes.  It is a pattern we use frequently and consistently in RCT - I counted about 5 times in RU, 10 times in RP.  It is set in the `stop()` method to signal thread termination.  This is actually the only reliable way in to terminate threads (in Python and any other pthread API), but it needs a collaborating thread (which we have fortunately).

src/radical/utils/zmq/pubsub.py
```diff
@@ -97,95 +43,77 @@ def uid(self):
...
+    def addr_pub(self):
+        return self._addr_pub
 
     @property
-    def type_out(self):
-        return 'SUB'
+    def addr_sub(self):
+        return self._addr_sub
+
+    def addr(self, spec):
```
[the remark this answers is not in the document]
andre-merzky: for `addr_sub` we have a property.  For `addr` we can't as it needs the `spec` parameter.  Am I missing something - not sure if that answered your question?

src/radical/utils/zmq/pubsub.py
andre-merzky: This has actually nothing to do with the heartbeat - this code applies to *any* ZMQ pubsub bridge.  The 500ms is the time we do a blocking wait (or rather `select`) on all sockets to check if messages arrive.  We need to interrupt now and then to check for `self._term` - this is the cooperation part from the discussion above :-)  If we don't do this, the thread would block forever and we would have no option to forcefully terminate it.

src/radical/utils/zmq/pubsub.py
andre-merzky: a pubsub topic is defined as a word in ZMQ - we just ensure that any given topic is indeed encoded as a single word.  This is only an internal representation - we don't use those topics for message filtering etc, so that encoding does not influence the other side of the channel.

src/radical/utils/zmq/pubsub.py
andre-merzky: You spotted that right: we say the channel can *either* serve callbacks, or the user can call `get() / get_nowait()` themselves.    It is not good to intermix both modes.  Assume you wait for a message of type 'A' followed by message of type 'B'.  You register a callback, and also do a blocking `get()` on the channel.  Now you have two competing `get()` operations on the socket, and there is no telling which one gets 'A', which one gets 'B', or which one gets both / none.  Either one will miss on 'A' and consider that an error.
This holds for any async communication: either you poll, *or* you push (here via callbacks) - mixing those will make clean protocols neigh impossible.

src/radical/utils/profile.py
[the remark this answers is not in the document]
iparask: Not really, I was just wondering

src/radical/utils/zmq/pubsub.py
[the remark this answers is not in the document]
iparask: Not really.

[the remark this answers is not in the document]
andre-merzky: not really missing something, or not really answered?  *confused*

tests/unittests/test_plugin_manager.py
```diff
@@ -50,15 +50,15 @@ def test_plugin_manager():
...
-        pass
+  # try:
+  #     pmgr = ru.PluginManager    ('troy')
+  #     plugin_strategy = pmgr.load('strategy', 'basic_late_binding')
+  #     import troy
+  #     session = troy.Session()
+  #     plugin_strategy.init_plugin(session)
+  #     plugin_strategy.execute    (None, None, None, None)
+  # except:
+  #     pass
```
iparask: If this is not used anymore, may I suggest to remove it completely?

tests/unittests/test_heartbeat.py
```diff
@@ -47,7 +47,7 @@ def proc():
 
         cb = get_counter()
 
-        def term():
+        def term_cb(uid):
```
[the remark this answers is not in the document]
andre-merzky: This is the signature the heartbeat manager expects those callbacks to have.  The test does not need the uid though.

tests/unittests/test_plugin_manager.py
andre-merzky: ack - fixed.

src/radical/utils/zmq/queue.py
```diff
@@ -265,59 +190,57 @@ def _bridge_work(self):
...
             buf = list()
-
-            i = -1
             while True:
 
-                i += 1
-
-                # we avoid busy pulling during inactivity
+                # check for incoming messages, and buffer them
+                ev_put = dict(no_intr(self._poll_put.poll, timeout=0))
```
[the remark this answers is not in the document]
andre-merzky: events on the `put channel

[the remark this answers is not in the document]
andre-merzky: > Why zero timeout? Wouldn't make sense to have a few milliseconds?

We do the sleep now in line 244.  The loop is now a busy loop as long as there are data to proxy, but slows down when no data are communicated on any of the channels.  Adding a timeout on all the select calls will cause the loop to slow on any idle channel even if the other channels are busy, and I wanted to avoid that.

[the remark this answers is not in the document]
andre-merzky: For the timeout question, there is also a comment around line 240.

src/radical/utils/heartbeat.py
```diff
@@ -34,48 +50,107 @@ def __init__(self, uid, timeout, interval=1, cb=None, term=None, log=None):
...
-                        self._term()
-
-                    os.kill(self._pid, signal.SIGTERM)
+                                       self._uid, uid, now, last, self._timeout)
+
+                    ret = False
+                    if  self._term_cb:
+                        ret = self._term_cb(uid)
+
+                    if ret is False:
```
andre-merzky: use None

src/radical/utils/heartbeat.py
```diff
@@ -34,48 +50,107 @@ def __init__(self, uid, timeout, interval=1, cb=None, term=None, log=None):
...
+                        # could not recover: abandon mothership
+                        self._log.warn('hb failure for %s - fatal (%d)', uid,
+                                self._pid)
+                        os.kill(self._pid, signal.SIGTERM)
+                        time.sleep(1)
+                        os.kill(self._pid, signal.SIGKILL)
+
+                    else:
```
andre-merzky: register new UID

andre-merzky:
Tests are passing locally now - but pep8 a nd pylint will take a while :-(

src/radical/utils/heartbeat.py
andre-merzky: done.

PR #196, opened 2019-12-16
Feature/mktar

andre-merzky:
Support tarball creation to simplify higher level codes (RP, RX)


andre-merzky:
@mturilli : do you have time for a quick review?  This optimizes repex (among others...)

src/radical/utils/misc.py
```diff
@@ -848,5 +850,46 @@ def rec_makedir(target):
             raise
 
 
+# ------------------------------------------------------------------------------
+#
+def mktar(tarname, fnames=None, data=None):
+    '''
+    Create a tarfile at the given `tarname`, and pack all files given in
+    `fnames` into it, and lso pack any `data` blobs.
```
mturilli: lso->also

src/radical/utils/misc.py
```diff
@@ -848,5 +850,46 @@ def rec_makedir(target):
...
+
+    `fnames` is expected to be list, where each element is either a string
+    pointing to a file to be added under that name, or a tuple where the first
+    element points again to the file to be packed, and the second element
+    specifies the name under which the file should be packed into the archive.
+
+    `data` is expected to
```
mturilli: spurious \n

src/radical/utils/misc.py
```diff
@@ -848,5 +850,46 @@ def rec_makedir(target):
...
+    be a list of tuples, where the first element is a set of bytes comprising
+    the data to be written into the archive, and the second element again
+    specifies the name of the tarred file.
+
+    Note that this method always create gzip'ed tarfiles, but will never change
+    the `tarname` to reflect that.
```
mturilli: Should bzip2 be the default compression seeing the (sometimes vast) superior efficiency?

src/radical/utils/misc.py
```diff
@@ -848,5 +850,46 @@ def rec_makedir(target):
...
+    '''
+
+    tar = tarfile.open(tarname, "w:gz")
+    if fnames:
+        for element in fnames:
```
mturilli: Should we check that the file indeed exist? Should we tell the user when it does not? I am thinking about unavoidable typos in the file name.

andre-merzky: the underlying tar module will raise a suitable exception already, so we don't need to replicate that test, fortunately.  I'll add a comment.

andre-merzky:
Thanks for the typos!  I changed the format to `bz2`.

PR #197, opened 2019-12-16
Feature/progress bar

andre-merzky:
add a progress bar to the reporter


mturilli:
Note that 3 Travis tests still fail. `src/radical/utils/reporter.py` seems in need of formatting attention, should we give some?

andre-merzky:
`.flake8` config was missing in RU, and is added now :-)  Thanks!

andre-merzky:
> Any reason not to use progressbar/2 or other libraries for this?

None specific, just considered it overkill to use an external tool - for now that progress bar is _really_ simple.  But thanks for the hint, should we ever expand on it we should reconsider!

PR #206, opened 2020-01-29
py3 miniconda installs gtod, but complains

andre-merzky:
revert to shell based gtod again, to avoid user confusion

### DIFF
--- a/bin/radical-utils-gtod
+++ b/bin/radical-utils-gtod
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+if test -z "$EPOCHREALTIME"
+then
+    python3 -c 'import time; print("%.6f" % time.time())'
+else
+    echo ${EPOCHREALTIME:0:17}
+fi
+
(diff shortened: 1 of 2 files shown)


iparask:
I am not sure what is this pull request supposed to do? Can you link to an issue or give me some context?

andre-merzky:
The PR avoids that `gtod` is compiled during installation.   We need some form of `gtod` (gettimeofday) to obtain subsecond resultion for profile events when the Python profiler is unavailable (like, in the bootstrapper or unit launch scripts).  That compilation leads to scary error messages even when the installation is successful.  I saw that on the Munich cluster during the saga tutorial, and it confused me, and the students even more.  I don't have the patience for a proper fix (and could not even reproduce this elsewhere), so I'd rather remove the compilation step and fall back to less precise / fast, but portable versions (bash or python).

iparask:
Okay. I agree with what you are saying,

PR #209, opened 2020-01-30
update README

### LINKED ISSUE #126: Readme.md links don't work
https://github.com/radical-cybertools/radical.utils/blob/devel/README.md

andre-merzky:
This also fixes #126


README.md
```diff
@@ -4,50 +4,32 @@
...
-* A [Thread wrapper](https://github.com/saga-project/radical.utils/blob/master/radical/utils/threads.py) (A thin convenience wrapper around threading.Thread)
-* A [Call tracer](https://github.com/saga-project/radical.utils/blob/master/radical/utils/tracer.py) (Support for call tracing / debugging)
-* A ['which' function](https://github.com/saga-project/radical.utils/blob/master/radical/utils/which.py)(Simliar to the 'which' command line tool)
-
-The package should in general be compatible with Python 2.6.
+This Python package contains shared code and tools for various [Radical
+Group](http://radical.rutgers.edu) projects. Sometimes we call it the Radical
+Kitchen Sink.  
 
+The package is compatible with Python 3.5 and later versions.
```
lee212: I personally find some issues using python 3.5 with minor patches, For example, 3.5.2 has an issue, https://github.com/python/typing/issues/259 while using hypothesis in unittest. Increasing the minimum requirement to python 3.6 would be simple but I see that Ubuntu Xenial 16.04 comes with python 3.5.2 default. What do you think?

mturilli: I think Python 3.5 is notorious for its issues so I would go for 3.6. I do not think Xenial support affects us too much.

PR #210, opened 2020-01-30
apply exception chaining

andre-merzky:
this fixes 23 - finally :-)


andre-merzky:
Pylinnt anf flake are annoying - can't silence both without writing more code than this PR needs in the first place...

PR #211, opened 2020-02-02
speed up uid generation

andre-merzky:
test: generating 1m simple IDs
before:
```shell
(ve3)  rivendell  merzky  ~/radical/radical.utils.devel  [optimize/generate_id *] $ time ./t.py
16.283207654953003
r:0m16.415s  u:0m14.472s  s:0m1.941s
```

after:
```shell
(ve3)  rivendell  merzky  ~/radical/radical.utils.devel  [optimize/generate_id] $ time ./t.py
3.7440223693847656
r:0m3.909s  u:0m3.877s  s:0m0.032s
```

runtime reduced to ~25%, but more importantly system load reduced to ~2% due to reduced number of hits on the file system.  This has a much larger impact on shared file systems.

### DIFF
--- a/src/radical/utils/ids.py
+++ b/src/radical/utils/ids.py
@@ -277,10 +277,10 @@ def _generate_id(template, prefix, ns=None):
     ret = template % info
 
     if '%(' in ret:
-      # import pprint
-      # pprint.pprint(info)
-      # print template
-      # print ret
+        # import pprint
+        # pprint.pprint(info)
+        # print(template)
+        # print(ret)
         raise ValueError('unknown pattern in template (%s)' % template)
 
     return ret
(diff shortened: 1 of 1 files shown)


src/radical/utils/ids.py
```diff
@@ -272,10 +283,6 @@ def _generate_id(template, prefix, ns=None):
       # print ret
         raise ValueError('unknown pattern in template (%s)' % template)
 
-    if 'client_notify' in ret and 'get' in ret:
```
mturilli: We don't need this anymore?

src/radical/utils/ids.py
```diff
@@ -272,10 +283,6 @@ def _generate_id(template, prefix, ns=None):
       # print ret
```
mturilli: Irrelevant: alignment?

src/radical/utils/ids.py
andre-merzky: Thanks for checking!  No, that was debug code which already should have been removed (was never triggered though).

src/radical/utils/ids.py
andre-merzky: thanks, fixed.

PR #217, opened 2020-02-18
Time class

iparask:
This is based on a slack discussion between @andre-merzky and me yesterday. I added a `sleep` method. This method clicks every `1/speed` seconds and checks the current time. There is there a thread pool of two threads that sleep for different amount of time concurrently and prints the correct values.

Let me know what you think.


src/radical/utils/custom_time.py
```diff
@@ -0,0 +1,67 @@
...
+    def sleep(self, amount):
+        '''
+        Sleeps for a specific amount of time units. Actual time spent is equal
+        to amount divided by speed.
+        '''
+        
+        now = self._seconds
+
+        while self._seconds < now + amount:
+            time.sleep(1 / self._speed)
```
andre-merzky: Why not just 
```python
def sleep(self, amount):
    time.sleep(amount . self._speed)
```
?

tests/unittests/test_time.py
```diff
@@ -0,0 +1,57 @@
...
+    t  = Time(tick=0.1, speed=3.0)
+    t0 = t.time()
+    time.sleep(3)
+    t.advance(60)
+    t1 = t.time()
+    t.stop()
+    assert(0.00 <= t0 <= 0.01), t0
+    assert(68.69 <  (t1 - t0) < 69.01), [t0, t1]
+
+def mysleep(amount, t_obj):
```
andre-merzky: Should this be a test?

tests/unittests/test_time.py
```diff
@@ -0,0 +1,57 @@
...
+    sleeping_thread2 = mt.Thread(target=mysleep, name='sleep2', args=(5,t,))
+    sleeping_thread1 = mt.Thread(target=mysleep, name='sleep1', args=(2,t,))
+    sleeping_thread1.start()
+    sleeping_thread2.start()
+    time.sleep(10)
+    sleeping_thread1.join()
+    sleeping_thread2.join()
+    t1 = t.time()
+    t.stop()
+    print(t1,t0)
```
andre-merzky: Should this also be a test?

iparask: Yes, along with `mysleep` which needs a name change.

src/radical/utils/custom_time.py
iparask: I'll change it. It still works.

PR #218, opened 2020-02-21
change bulk defaults

andre-merzky:
This increases default bulk size to 1024, and switches to `0` to trigger the default value.


andre-merzky:
@mtitov : Oh bugger, I saw the comment after merge...  Yes, that makes sense, will do.

PR #220, opened 2020-02-25
silence queue logs

andre-merzky:
This fixes radical-cybertools/radical.pilot/issues/2085

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -40,4 +40,6 @@ ignore          =
      # unused argument
      W613,
      # ambiguous var name (to whom?)
-     E741
+     E741,
+     # explicit delegation
+     W235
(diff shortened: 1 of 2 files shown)


mturilli:
Note that one of the test fails with:
```
src/radical/utils/zmq/pubsub.py

  32  4   W0235   Useless super delegation in method '__init__' (useless-super-delegation)
```
Is that the case?

andre-merzky:
> Note that one of the test fails with:
> 
> ```
> src/radical/utils/zmq/pubsub.py
> 
>   32  4   W0235   Useless super delegation in method '__init__' (useless-super-delegation)
> ```
> 
> Is that the case?

Kind-of yes, as the default inheritance will kick in anyway.  But the explicit delegation makes this, well, explicit, so it's there on purpose.  I disable the warning.

andre-merzky:
> Could you merge this also in hotfix/frontera so that I can measure how relevant the reduction in log size is?

@mturilli : since the fix is in RU, the `hotfix/frontera` RP branch should benefit immediately.

andre-merzky:
> Could you merge this also in hotfix/frontera so that I can measure how relevant the reduction in log size is?

@mturilli : since the fix is in RU, the `hotfix/frontera` RP branch should benefit immediately.

Issue #222, opened 2020-02-28
Name for cnt-file (inside method generate_id) is defined incorrectly


mtitov:
Example: `~/.radical/utils/rp.session<…>/ru_matitov_pilot.%(item_counter)04d.cnt`
(code where name is defined is [here](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/ids.py#L261))

mtitov:
BTW is the purpose of `~/.radical` dir to keep configs? should such files (as above) be kept at `sandbox` dir?

andre-merzky:
I seem to be missing the point of the ticket (and of the PR #223), apologies @mtitov !  Can we restart the discussion, please?  So, why do you think that this is incorrectly defined - what exactly is broken?  Do you have a script to reproduce the problem?

> BTW is the purpose of ~/.radical dir to keep configs? should such files (as above) be kept at sandbox dir?

The `~/.radical` dir has two purposes: to have a place for the user to drop config files which overwrite or extend the configs provided by the RCT installation tree, and to keep state between runs (like counters and job information etc).  Neither of those would work feasible in a sandbox, as sandboxes are expected to be temporary and might disappear at any time.

mtitov:
sure! agree to keep the discussion here

> So, why do you think that this is incorrectly defined - what exactly is broken?

just running the example `00_getting_started.py`, it creates files (see below) in the dir `/Users/matitov/.radical/utils/rp.session.MT-MBP-X2.local.matitov.018322.0000/`:
```
ru_matitov_pilot.%(item_counter)04d.cnt
ru_matitov_pmgr.%(item_counter)04d.cnt
ru_matitov_umgr.%(item_counter)04d.cnt
ru_matitov_unit.%(item_counter)06d.cnt
```
the code that is responsible for it is [here](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/ids.py#L261)

adding here my previous reply:
> I was a little confused with possible original purpose of it: either 1) always change the name with the current `item_counter` (read the current `item_counter`, delete corresponding file, and then create a new file with new `item_counter` in name and as a text inside); or 2) create a set of files while new value for `item_counter` appears (I understand that this option is less probable, but there is no deletions for the 1st option, and without knowing the value it wouldn’t be possible to read the file at the beginning).
> For now cannot come up with something more elegant. 

Also (from the previous reply) there is a proposal for small changes:
[...]

PR #223, opened 2020-02-28
Name for cnt-file (inside method generate_id) is defined incorrectly

mtitov:


mtitov:
@andre-merzky I was a little confused with possible original purpose of it: either 1) always change the name with the current `item_counter` (read the current `item_counter`, delete corresponding file, and then create a new file with new `item_counter` in name and as a text inside); or 2) create a set of files while new value for `item_counter` appears (I understand that this option is less probable, but there is no deletions for the 1st option, and without knowing the value it wouldn’t be possible to read the file at the beginning). 

For now cannot come up with something more elegant. I would think that this case (with `item_counter` and `ru.ID_CUSTOM`) could be reworked (at least to have a separate parameter `new_template=None` if `mode==ID_CUSTOM`, so not to mix with `prefix` parameter?), but there are many usages of it

p.s. also need some comments/opinions about `.radical` dir at the issue #222

andre-merzky:
@mtitov : so, what are we going to do with this one?  I know I approved it, but the premise of the edit was a misunderstanding.  It does make the names a bit more readable (no placeholders), but at the same time a bit less correct (no placeholders ;-) ).   I would be inclined to close the PR unmerged, and instead consider moving to a csv or json file for maintaining counter state (after checking performance overhead).  Any opinion?

mtitov:
@andre-merzky agree that previous update didn't fulfill the expected outcome (it did decrease the uniqueness), so I've made changes in `prefix` accordingly - changed that "template"-format field to just a name of parameter, thus file name would look like `ru_matitov_pmgr.item_counter.cnt`, if prefix was like `pmgr.%(item_counter)04d`. Also agree to keep counters in different way, but considering to have a plain name :)

I've updated test file as well - it does test these files with counters

src/radical/utils/ids.py
```diff
@@ -210,7 +208,7 @@ def _generate_id(template, prefix, ns=None):
 
     state_dir = _BASE
     if ns:
-        state_dir += '/%s' % ns
+        state_dir = os.path.join(_BASE, ns)
```
andre-merzky, line 211: The only reason to use `path.join` really is Windows support, which we will never have, so don't bother.  We do path manipulations all over the place, and while `path.join` is recommended usually, it simply does not add any value and makes the manipulation less transparent (one needs to be more careful with the joined paths which must not start with a slash).

src/radical/utils/ids.py
```diff
@@ -250,22 +248,33 @@ def _generate_id(template, prefix, ns=None):
...
         os.write(fd, line)
         os.close(fd)
 
     if '%(item_counter)' in template:
-        fd = os.open("%s/ru_%s_%s.cnt" % (state_dir, user, prefix),
-                                          os.O_RDWR | os.O_CREAT)
+
+        # clean up "prefix" to use in file name
+        #  FIXME: extend same procedure for other cases (with regex?)
+        if '%(item_counter)' in prefix:
```
andre-merzky, line 268: I like the scheme, that makes sense.  Unintended collisions are indeed very unlikely.  
To generalize, maybe something like this:
```python
for k in info:
    prefix = prefix.replace('%%(%s)s' % k, k)
```
?

andre-merzky: Quick note that we may want to quickly profile this.  For large runs, we may want to create a million task IDs (we are slowly getting into that range).  The code is not the fastest anyway (yet another reason to consider alternatives), but we should a oid adding too much.   I assume this is much smaller than the FS overhead though.

tests/unittests/test_ids.py
```diff
@@ -14,63 +21,59 @@ def test_ids():
...
-    assert (isinstance (id_b_2, str)), 'isinstance("%s", str)' % (id_b_2)
-    assert (isinstance (id_c_1, str)), 'isinstance("%s", str)' % (id_c_1)
-    assert (isinstance (id_c_2, str)), 'isinstance("%s", str)' % (id_c_2)
-    assert (isinstance (id_c_1, str)), 'isinstance("%s", str)' % (id_c_1)
-    assert (isinstance (id_c_2, str)), 'isinstance("%s", str)' % (id_c_2)
-    assert (isinstance (id_d_1, str)), 'isinstance("%s", str)' % (id_d_1)
-    assert (isinstance (id_e_1, str)), 'isinstance("%s", str)' % (id_e_1)
+    id_a_1 = ru.generate_id('a')  # default mode is ID_SIMPLE
+    id_a_2 = ru.generate_id('a', mode=ru.ID_SIMPLE)
+    assert(id_a_1 == 'a.0000'), 'id_a_1 != a.0000'
```
andre-merzky, line 26: removing the `%s` replacement from the assert will then not report the actual value on errors.  Is that intentional?

tests/unittests/test_ids.py
```diff
@@ -14,63 +21,59 @@ def test_ids():
...
+    with open(fname) as fd:
+        day_counter = int(fd.readline())
+    id_d_2 = ru.generate_id('d.%(day_counter)06d', mode=ru.ID_CUSTOM)
+    assert(id_d_2 == 'd.%06d' % day_counter), 'id_d_2 != d.%06d' % day_counter
+
+    ns = 'id_e'
+    _ = ru.generate_id('e.%(item_counter)04d', mode=ru.ID_CUSTOM, ns=ns)
+    # file format: /tmp/.radical/utils/id_e/ru_<user>_<cleaned_prefix>.cnt
+    fname = os.path.join(ru_base_dir, ns, 'ru_%s_e.item_counter.cnt' % user)
+    assert (os.path.isfile(fname)), 'file with <item_counter> does not exist'
```
andre-merzky, line 65: Hmm, I find the tests less informative and a bit more convoluted now, but I guess it increases coverage?  Maybe simply let the original tests in place, too?

andre-merzky:
This looks all very reasonable, thanks!  I sprinkled a couple of comments throughout the code, let me know what you think.

src/radical/utils/ids.py
mtitov: I still would encourage to use it (definitely not to go through the code right away to make changes, but whenever it is possible). Such approach also helps to eliminate double-slashing (`//`), when the base-dir ends with `/` and then dir/file is added as `/<path>`, it as well accepts [path-like objects](https://docs.python.org/3/glossary.html#term-path-like-object), even we don't use it now, for later it might make updates easier. In general I would like to avoid "manual"-operations if it is possible (as here for paths creation)

src/radical/utils/ids.py
mtitov: even better is to use regular expression here (e.g., `re.sub(r'\%\((.*)\).*', r'\1', '%(test)04d')`). 

But in general that custom prefixes are tricky, they still uses only that variables that are defined inside `_generate_id`, in this case may be better to create corresponding templates inside `ids.py` whenever there is a request for a new id-format. I would prefer to avoid template-like `prefix` parameter since it still requires to know the internal parameters and it loses the meaning of `prefix` (so this example may be like `ru.generate_id('pmgr', mode=ru.ID_ITEM_COUNTER)`)

tests/unittests/test_ids.py
mtitov: I didn't find much meaning in that, since the string just brings the alert message, so the current string would say that in our id-variable the value is not what we expect (and then extra info is provided by test system)

tests/unittests/test_ids.py
mtitov: in original tests there were too many duplications (e.g., compare `id_x_x` with a certain string value, and then check if `id_x_x` is of string type and/or compare with others while they also were already checked with specific string values) that's why I thought just to go with specific parts to check/test

> I find the tests less informative

they are actually splitted into the groups, so it is easy to track which mode or what inside the certain mode is checked. It might look so just at the first :)

mtitov:
@andre-merzky thank you! BTW we can already re-think [this](https://github.com/radical-cybertools/radical.utils/pull/223#discussion_r413937263) (the second part of the comment), may be inside RP - such template is used only 4 times, so we can push it here, and slowly extend for other RCT components

PR #227, opened 2020-03-22
Hotfix/atfork zmq

andre-merzky:
When setting `export RADICAL_UTILS_PATCHATFORK=true`, the PubSub Subscriber and Queue Listener will clean out registered callbacks and listener thread handles on `os.fork`, so that the new process can instantiate new callback handlers.  

This is a stopgap solution to help RE to function (radical-cybertools/radical.entk/issues/428).  In general RU will make no promises to keep ZMQ connectivity valid across process forks.  The process management in RE should be reconsidered in that context.


andre-merzky:
No idea how to turn off the remaining code style errors - one is wrong, the others all have an (ugly) `# noqa` flag which seems to be ignored for some reason.

PR #230, opened 2020-04-01
Fix/listener termination

andre-merzky:
This is again related to radical-cybertools/radical.pilot/pull/2111 and implements some explicit thread cleanup on RU.ZMQ level.  However, this PR is a bit larger, as I originally based the work on the `project/covid-19` branch which also changed code on the ZMQ layer to better support the master/worker communication in RP's task overlay.  Now, those changes are not easily separated, so this PR includes both sets.  On the upside, this PR is not necessary for resolving the pmgr.launching zombie apocalypse, and has thus sufficient time for reviews.


src/radical/utils/zmq/utils.py
```diff
@@ -0,0 +1,149 @@
+
+import os
+import zmq
+import errno
+import msgpack
+
+import radical.utils as ru
```
iparask, line 7: Why is this cyclic import happening?

andre-merzky: I don't quite understand why this should be a problem:
```
$ python3 -c 'import radical.utils'
$ python3 -c 'import radical.utils.zmq'
$ python3 -c 'import radical.utils.zmq.utils'
$ python3 -c 'import radical.utils as ru; q = ru.zmq.Queue(channel="foo"); print(q)'
<radical.utils.zmq.queue.Queue object at 0x7fe28a293f50>
$ 
```

Let me look into the travis complains a bit more closely...

andre-merzky: Hmpf, locally all tests run (with one exception due to logging, will fix):
```
$ pytest -qqq tests
...........................................................................             [100%]
```

Any idea?  I mean, I can work around it, but I don't like when tests want me to change code and I don't understand why...

andre-merzky: Ok, I just changed, whatever.  About the other errors: 
  - I silenced the one about the unused log method.  
  - The ones about private access to `_callbacks` in the `at_exit` handlers are already marked `noqa`, which seems to be ignored.  
  - The pylint errors in the tests about unused `ru` is simply wrong: if you look in the code, `ru` is being used in the metaclass specifier.  I don't know how else to write the code, it is the only way to express metaclasses in py3.
  - the ones in the tests about accessing `_foo` also have a `noqa` tag which seems to be ignored.

iparask: I also just run the tests locally and passed. I still do not understand the reason for importing the whole package in a submodule of the package. It creates a continuous cyclic imports that may result to problems. Is this correct or do I understand something wrongly?

andre-merzky: It should be a dead end: zmq.utils is not imported in zmq.__init__.py, so should not be pulled when the module structure is parsed.  Well, maybe the way the parser works is version dependent or whatever, it's probably not worthwhile tracking this down...

src/radical/utils/zmq/bridge.py
```diff
@@ -88,6 +26,9 @@ def __init__(self, cfg):
                                level='DEBUG', path=self._cfg.path)
         self._prof    = Profiler(name=self._uid, path=self._cfg.path)
 
+        if 'hb' in self._uid or 'heartbeat' in self._uid:
+            self._prof.disable()
```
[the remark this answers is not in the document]
andre-merzky: This disables the profiler for heartbeat bridges: those messages are not useful to profile and add a lot of volume and noise.  Unexpected heartbeat events are still profiled in their respective callbacks.

andre-merzky:
Thanks @mturilli , and no worries.

andre-merzky:
The main change is the implementation of `_stop_listener` in both pubsub and queue receivers (Getter,  Subscriber).  That requires some changes to how the listener threads are maintained, how they are associated with their `term` signal, etc.

The changes related to master/worker are to implement a listener on the queue Getter class in the first place - that mechanism only existed in the pubsub Subscriber before.  Not needing to do that thread management simplifies both master and worker a lot (both listen on different queues).

I understand that the PR is not very clean, really sorry for that.  But that's also why I opened it: it's not urgent yet, so we have time for review, and it's better to do now before it collects even more commits (master/worker are in active use now).

Let me know if you have any suggestion to help with the review!  

Thanks, Andre.

andre-merzky:
@ping - any  suggestion on how to proceed?

andre-merzky:
@iparask : Ioannis, let me put you on the spot (comes with the job :-P)  Can we have a quick call today to go through this one and make sure you are happy with it, so it can be merged?  Would like to get it out...

andre-merzky:
@iparask : changed cb registration as discussed.

iparask:
I'm okay with the PR as we discussed. I just realized that a test fails. Do you think you can fix it before merging?

If the test is fixed, I think the PR can be merged.

andre-merzky:
> I just realized that a test fails. Do you think you can fix it before merging?

Haha - the test fails because it tests multiple callbacks, which we just disabled :-P  I'll disable it.

PR #233, opened 2020-04-21
Fix/rlocks

andre-merzky:
As discussed - @iparask : did that change work as expected?


andre-merzky:
@andre-merzky : check if a test exists

andre-merzky:
Test needs to be written

andre-merzky:
added tests

PR #235, opened 2020-05-10
Add configurable csv field size limit

mturilli:
This should enable profiling of very large sessions.

### DIFF
--- a/src/radical/utils/profile.py
+++ b/src/radical/utils/profile.py
@@ -52,6 +52,11 @@ NTP_DIFF_WARN_LIMIT = 1.0
 # cache the result.  We use a disk cache which is valid for 1 minute
 NTP_CACHE_TIMEOUT = 60  # disk cache is valid for 60 seconds
 
+# maximum field size allowed by the csv parser.  The larger the number of 
+# entities in the profile, the larger the size of the filed required by the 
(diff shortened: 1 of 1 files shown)


mturilli:
About the failing tests, can we make `'_foo' : str,` in `tests/unittests/test_description.py:_schema` `'foo' : str,`, seeing that `_schema` is already private?

mturilli:
About the failing test, in  I see:
```
try:
    # make pylint happy - this is optional code
    from Xlib import X, display, Xutil
except:
    raise RuntimeError('need Xlib module to work')
```
I am not sure how this should make pylint happy and it doesn't. Should we use `pip install python3_xlib` instead and, seeing that we want to use Xlib (I don't, BTW) add python3_xlib as a required module?

andre-merzky:
Xlib: well, it used to pacify pylint :-P  xlib is not always available nor installable, I'm not sure if we want to make the tests depend on it.  Not sure how to handle that best.  Maybe installing it in travis is worth a try, we'll see if it's installable there...

andre-merzky:
`_foo`: accessing `_foo` this way *is* the test.  I don't know why the `noqa` tag is ignored, and gave up trying to understand :-/

PR #237, opened 2020-05-12
var consistency

andre-merzky:
Two different versions of the env var crept into our code / use, and this fix allows for that.  In the long run, `RADICAL_BASE_DIR` should be retired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
     - COVERAGE=false
     - PYTEST="coverage run -m pytest -ra --timeout=600 -vvv --showlocals"
     - RADICAL_DEBUG=True
+    - RADICAL_DEBUG_HELPER=True
 
   matrix:
     - MAIN_CMD="$PYTEST tests/"
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
     - CODECOV_TOKEN="18d4cc7a-d35c-44ec-9684-e6f2ec7702fc"
     - COVERAGE=false
     - PYTEST="coverage run -m pytest -ra --timeout=600 -vvv --showlocals"
+    - RADICAL_DEBUG=True
 
   matrix:
     - MAIN_CMD="$PYTEST tests/"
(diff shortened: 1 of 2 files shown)


andre-merzky:
Because of something else - the tests now need `RADICAL_DEBUG` to be set, but the `os_environ` directove I added to the lock tests is too late.  I'll fix the travis setup.

andre-merzky:
Any idea why that doesn't work?  I don't see the env setting applied...

iparask:
The env setting is applied (see [here](https://travis-ci.com/github/radical-cybertools/radical.utils/jobs/332542389#L185)). Also, I setup locally the variable before calling purest and printed the variable value. This is what I got:

```

tests/unittests/test_locks.py::test_locks FAILED                                                        [100%]

================================================== FAILURES ===================================================
_________________________________________________ test_locks __________________________________________________

    def test_locks():
        '''
        Test debug lock wrappers
        '''

        print(os.environ['RADICAL_DEBUG'])
        os.environ['RADICAL_DEBUG'] = 'True'
        print(os.environ['RADICAL_DEBUG'])
        l  = ru.Lock()
        rl = ru.RLock(name='bar')

        assert(not l.waits)
        assert(not rl.waits)

        with l:
            with rl:
                assert(not l.waits)
                assert(not rl.waits)

>       assert(l.name  in ru.debug._debug_helper.locks)                       # noqa
E       AttributeError: 'NoneType' object has no attribute 'locks'

tests/unittests/test_locks.py:29: AttributeError
-------------------------------------------- Captured stdout call ---------------------------------------------
True
True
```

Is there a chance that the why the `_debug_helper` is initialized cases issues and not the variable?

andre-merzky:
>  Is there a chance that the why the _debug_helper is initialized cases issues and not the variable?

Sorry, not sure I understand the question.  Can you rephrase that maybe?

Either way: I forgot that somebody requested a separate env var for the debug helper.  I added that to the travis setup now.

iparask:
Okay, I see. Does this mean that there are too many environment variables?

andre-merzky:
> Okay, I see. Does this mean that there are too many environment variables?

the debug helper used to be enabled on `RADICAL_DEBUG`.  I can't remember why exactly that was changed, but I do remember a request for that.  Now, if that's too many env vars or not is in the eye of the beholder I guess - I wouldn't have minded to stick to one `DEBUG` env....

PR #240, opened 2020-06-01
remove obsolete code, increase coverage

andre-merzky:


mturilli:
I see in the test failures, the following: 
```
>       assert(data['C'].count('A') + data['C'].count('B') +
               data['D'].count('A') + data['D'].count('B') == c_a + c_b)
E       assert 292 == 300
E         -292
E         +300
```
How is it supposed to pass?

PR #242, opened 2020-06-09
Feature/config with origins

andre-merzky:
Opened PR for review and discussion.


andre-merzky:
@mtitov : is it ok to close this in favor of #243 ?

mtitov:
@andre-merzky would we need nested-config function from there? https://github.com/radical-cybertools/radical.utils/blob/feature/config_with_origins/src/radical/utils/config.py#L146 (for some input config files) that the only part I would go for to save for later

mtitov:
@andre-merzky yeah, agree to close it (couldn't to come up with possible use cases and since json-configs will be converted into Munch-based ones, then approach of this PR is obsolete) p.s. deleted previous message to avoid any more confusion

PR #243, opened 2020-06-12
Feature/config inheritance

andre-merzky:
This consolidates the Munch base classes used in RU.  `ru.Description` used it's own Munch implementation which scaled better than `munch.Munch` (description is used for task descriptions, and we can have many of those, so performance and memory usage is important here).  `ru.Config` used the plain `munch.Munch`.  This PR consolildates both to `ru.Munch`.  This now allows us to specify schemas and defaults for `ru.Config` classes.

PS.: the diffs might be easier to read for the individual commits, specifically for the `ru.Munch` class which moved into it's own source file.


andre-merzky:
Thanks for the commit!  As for `super`: I don't have a strong opinion: `Munch` is more explicit / reradable IMHO, `super` probably somewhat cleaner, so I don't mind either way. Feel free to change if you have that preference!

PR #247, opened 2020-07-10
Extend set of parameters for MongoDB connection

mtitov:


mtitov:
New possible url format is `"mongodb+ssl://<username>:<password>@<host>:<port>/<db>?tlsAllowInvalidCertificates=True"`

mtitov:
> do you need me to?

@mturilli no, that's not necessary, I did test it at Lassen (with and without that extra parameter), thanks!

mtitov:
@andre-merzky agree with keeping it at one place (`split_dburl`), and have a proposal about what that function could return. It is possible that more optional parameters would be needed, thus could we do the following:
```python
def split_dburl(dburl, default_dburl=None):
...
    query_options = {'ssl': False}

    if 'ssl' in url.schema.split('+'):
        query_options['ssl'] = True
        url.schema = 'mongodb'
...
    if url.query:
        from urllib.parse import parse_qsl
        # convert query format from str to dict
        _query = dict(parse_qsl(url.query))
        # control of which options are transferred
        from distutils.util import strtobool
        query_options['tlsAllowInvalidCertificates'] = \
            bool(strtobool(_query.get('tlsAllowInvalidCertificates', 'False')))
...
return [host, port, dbname, cname, pname, user, pwd, query_options]

def mongodb_connect(dburl, default_dburl=None):
...
    [host, port, dbname, cname, pname,
           user, pwd,    options] = split_dburl(dburl, default_dburl)

    mongo = pymongo.MongoClient(host=host, port=port, **options)
...
```
so, whenever new optional parameter will be added, then we will just update `query_options` at `split_dburl`.

andre-merzky:
I know that premature optimization is the root of all evil - but could we not use an external function to check if a string is true?  
```python
bool(query.get("...", 'False').lower() in ['true', '1']))
```
should do the trick (and will also not raise an exception on invalid values but use the default).  Otherwise that looks great...

Hmm, I always  seems to have _juuuust one small thing_ complain about, ey?  Need to ponder why that is... :-P

mtitov:
thanks @andre-merzky! will update accordingly

PR #249, opened 2020-07-15
line buffering is not multiprocess-safe for high-frequency writes

andre-merzky:
This is troublesome, as we introduced line buffering to avoid incomplete and/or corrupted profiles.  I did not yet manage to find a viable mode to flush after lines in a multiprocess envionment.  Nevertheless, this fix is needed as it triggers a hangup in the task overlay worker when running at full speed (even for small resources sets).

### DIFF
--- a/src/radical/utils/profile.py
+++ b/src/radical/utils/profile.py
@@ -52,8 +52,8 @@ NTP_DIFF_WARN_LIMIT = 1.0
 # cache the result.  We use a disk cache which is valid for 1 minute
 NTP_CACHE_TIMEOUT = 60  # disk cache is valid for 60 seconds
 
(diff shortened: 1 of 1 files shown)


andre-merzky:
PS.: I don't have a decent stand-alone reproducer for the problem - but the TO worker reliably hangs when line-buffering is used for fast workloads.  Took me ages to track down to the profiler's I/O layer... :-/

andre-merzky:
Yes, we may end up with truncated profiles - but writes up to 1024 bytes should be atomic under POSIX file systems, so lets hope for the best.  FWIW, unbuffered I/O for string writes is, for whatever reason, not supported by python (see [python ticket](https://bugs.python.org/issue17404) - open since 2012, so I don't expect that to be fixed anytime soon).

PR #250, opened 2020-07-16
add script to function converter

andre-merzky:
The added method (`ru.script_2_func`) will read a python script and return a function handle which, wenn called, will run the content of the python script within the current interpreter.  The expected use case is the WF0 function executor which will use that method to execute external autodock scripts.

### DIFF
--- a/src/radical/utils/misc.py
+++ b/src/radical/utils/misc.py
@@ -1018,10 +1018,12 @@ def script_2_func(fpath):
         global _RU_stdout
         global _RU_stderr
         global _RU_except
+        global _RU_exit
 
         _RU_stdout = None
         _RU_stderr = None
         _RU_except = None
+        _RU_exit   = None
 
         tmp_code = ''.join(prefix)   \
                  + ''.join(code_lines) \
--- a/src/radical/utils/misc.py
+++ b/src/radical/utils/misc.py
@@ -961,9 +961,86 @@ def script_2_func(fpath):
           current Python interpreter.
     '''
 
+    prefix  = []
+    postfix = []
+
+
+    with open(fpath, 'r') as fin:
+        code_lines = fin.readlines()
+
+    # ensure that 'if __name__ == '__main__' works
+    prefix  += '__name__ = "__main__"\n\n'
+
+    # capture globals
+    prefix.append('global _RU_stdout\n')
+    prefix.append('global _RU_stderr\n')
+    prefix.append('global _RU_except\n')
+
+    # redirect stdout and stderr to be captured in global string vars
+    prefix.append('from io import StringIO\n')
+    prefix.append('_RU_oldout  = sys.stdout\n')
+    prefix.append('_RU_olderr  = sys.stderr\n')
+    prefix.append('sys.stdout  = _RU_stdout = StringIO()\n\n')
+    prefix.append('sys.stderr  = _RU_stderr = StringIO()\n\n')
+
+    # wrap the code in a try/exec to capture exit codes and failures
+    prefix.append('_RU_except = None\n')
+    prefix.append('try:\n')
+
+    postfix.append('except Exception as e:\n')
+    postfix.append('    _RU_except = str(e)\n')
+
+    postfix.append('sys.stdout = _RU_oldout\n')
+    postfix.append('sys.stderr = _RU_olderr\n')
+
+    # examine _RU_stdout.getvalue()
+
+    # indentation of all code lines which go into the try/except
+    code_lines = ['    ' + line for line in code_lines]
+
+    # create closure which can be used to call the code
+    def ret(*argv):
+
+        if len(argv) == 1 and isinstance(argv[0], list):
+            argv = list(argv[0])
+
+        # if args are given, ensure that sys.argv gets set
+        if argv:
+            # indent - this is within try/catch
+            prefix.append('    import sys\n')
+            prefix.append('    sys.argv = ["dummy"] + %s\n' % str(argv))
+
+        global _RU_stdout
+        global _RU_stderr
+        global _RU_except
+
+        _RU_stdout = None
+        _RU_stderr = None
+        _RU_except = None
+
+        tmp_code = ''.join(prefix)   \
+                 + ''.join(code_lines) \
+                 + ''.join(postfix)
+
+        # exec the resulting code, ensure to pass globals
+        exec(tmp_code, globals())
+
+        return _RU_stdout.getvalue(), _RU_stderr.getvalue(), _RU_except
+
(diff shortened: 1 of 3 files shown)


mtitov:
@andre-merzky it does miss constants definition at the beginning of script-file, e.g.:
```
TMP_VAR = 'test'

def print_var():
    print(TMP_VAR)

if __name__ == '__main__':
    print_var()
```
it will give a `NameError`

I have another proposal of converting text into module:
```
def script_2_func(fpath):

    import importlib.machinery
    import importlib.util
    loader = importlib.machinery.SourceFileLoader('__main__', fpath)
    spec = importlib.util.spec_from_loader(loader.name, loader)
    mod = importlib.util.module_from_spec(spec)

    def ret(*argv):
        import sys
        sys.argv = ['dummy'] + list(argv)
        loader.exec_module(mod)
    return ret
```

p.s. btw in test there should be `func('foo')` or `func(*['foo'])`

andre-merzky:
Hmm - I don't see why the constant definition would not work...  I added it to the test, and it seems ok?  Do you have a reproducer for the problem?  

As for importlib: I don't trust Python's module management much :-)  But yeah, that would also work.

mtitov:
When there is a constant variable, which is used in function, and then that function is called in `if __name__=="__main__"` section:
```
    os.write(tmpfile, ru.as_bytes("""#!/usr/bin/env python3
BAR = 'bar'
def get_bar():
    return BAR
if __name__ == '__main__':
    import os,sys
    os.system('echo "%%s %%s OK" >> %s' %% (sys.argv[1], get_bar()))
""" % tmpname))
```
gives
```
>   ???
E   NameError: name 'BAR' is not defined

<string>:9: NameError
```

p.s. in general I'm suspicious with current implementation due to possible conflicts because of namespaces and how objects are registered there (above is one of the possible examples, don't know will we have more), that's why I thought of `importlib`. And yeah, importing is always tricky :)

andre-merzky:
Good catch, thanks!  Ok, importlib it is.  One only drawback I see that calling the handle changes `sys.argv` for the current process - I think we can live with that, and I added a note in the doc.

src/radical/utils/misc.py
```diff
@@ -8,6 +8,7 @@
 import tarfile
 import datetime
 import tempfile
+import importlib
```
mtitov, line 11: `pylint` might give an error here, better to use:
```
import importlib.machinery
import importlib.util
```

src/radical/utils/misc.py
```diff
@@ -953,31 +954,19 @@ def script_2_func(fpath):
 
         my_func = ru.script_2_func('/tmp/my_script.py')
         my_func('-f foo -b bar'.split())
```
mtitov, line 956: `my_func(*'-f foo -b bar'.split())` (add `*`)
and maybe also update last couple sentences in function description? (about `exec`)

mtitov:
@andre-merzky thank you! added comments inline, but overall looks good!

src/radical/utils/misc.py
andre-merzky:
```
 $ pylint src/radical/utils/misc.py 

------------------------------------
Your code has been rated at 10.00/10

```
What do you see?  Why would that import be incorrect?

mtitov: inside `importlib.__init__.py` there is following:
```__all__ = ['__import__', 'import_module', 'invalidate_caches', 'reload']```

File: src/radical/utils/misc.py (hunk 2 above)
Comment (andre-merzky) [reply src/radical/utils/misc.py]: Thanks, update pushed (to allow both versions).

File: src/radical/utils/misc.py (hunk 1 above)
Comment (andre-merzky) [reply src/radical/utils/misc.py:11]: I am sorry, I don't understand - is pylint compaining about the content of importlib then?  That should not concern us, really?

Comment (mtitov) [reply src/radical/utils/misc.py:11]: PyCharm inspection shows this to me, it is with a warning only, so both ways will work. I showed just how that package is organized, kinda import of sub-packages directly should be made, since top level of `importlib` provides only 4 functions. But, I'm not pushing this, just as a recommendation

**Comment by andre-merzky** [comment]
Hey @mtitov : I reverted the code to using `exec`, as that allows to change the code in question.  The constants problem can be solved by passing `globals()`, fortunately.  The code now captures stdio and raised exceptions, which is what I'll need for wf0.  If you see an easy way to achieve that with importlib, happy to reconsider, and also don't mind having two versions: one with and one w/o capture.

**Comment by mtitov** [comment]
@andre-merzky sorry, looks overcomplicated to me... I like the idea to preserve std-out/-err values, but don't think it is right to catch exceptions for user, since it is up to user either to catch an exception or let it be raised, as it is always with functions (I see that certain info about exceptions is provided to user, but it is limited, since it is exception names and codes, but not exception objects). BTW at least one exception is not considered here - `KeyboardInterrupt` (that's why I'd like keep it for user). I see this method as a "convertor", thus would propose just extend the previous approach as following (`func` will return only `stdout` and `stderr`)

```
    loader = importlib.machinery.SourceFileLoader('__main__', fpath)
    spec   = importlib.util.spec_from_loader(loader.name, loader)
    mod    = importlib.util.module_from_spec(spec)

    def ret(*argv):
        _saved_sys = sys.argv, sys.stdout, sys.stderr
        from io import StringIO
        _sys_std = [StringIO(), StringIO()]
        sys.stdout, sys.stderr = _sys_std
        if len(argv) == 1 and isinstance(argv[0], list):
            argv = argv[0]
        sys.argv = ['dummy'] + list(argv)
        loader.exec_module(mod)
        output = _sys_std[0].getvalue(), _sys_std[1].getvalue()
        sys.argv, sys.stdout, sys.stderr = _saved_sys
        return output

    return ret
```

and work with it accordingly with updated tests, here is just parts that would be changed:
```
    with pytest.raises(ValueError):
        out, err = func('foo bar'.split())
        assert(out == 'hello\n')
        assert(err == 'world')
...
    with pytest.raises(SystemExit):
        out, err = func('foo', 'exit')
        assert(out == 'hello\n')
        assert(err == 'world')
```

andre-merzky:
Hehe, our definition of compicated differs :-D  

Ack on the user responsibility for the exception handling - but since in the given case I have to run 3rd party python scripts, the added exception handling (and the exit code) helps a lot with debugging.  Passing a fullexception object also works, and adding KeyboardInterrupt makes sense.

Also, changing stdio externally to the call makes sense, good point.

Lets me sit on this for a while - it's getting usable for wf0 now, and I need to make pogress on that end.  Thanks for the great feedback!

PR #254, opened 2020-07-29
Min python version 3.6

mtitov:


bin/radical-stack-clone
```diff
@@ -127,7 +127,7 @@ fi
 
 if test -z "$stack$branch$tag"
 then
-    usage 'missing stack file or branch'
+    usage "missing stack file or branch"
 fi
```
andre-merzky, line 131: single quotes are correct if no expansion is wanted?

bin/radical-utils-scheduler.py
```diff
@@ -13,8 +13,8 @@
 try:
     # make pylint happy - this is optional code
     from Xlib import X, display, Xutil
-except ImportError as e:
-    raise RuntimeError('need Xlib module to work') from e
+except ImportError as exc:
+    raise RuntimeError('need Xlib module to work') from exc
```
andre-merzky, line 18: Not sure I understand this change - `e` should be a valid identifier here?

bin/radical-stack-clone
mtitov: yeah, small update just to keep the consistency with other strings in this file

bin/radical-utils-scheduler.py
mtitov: because of pylint :) it didn't like that `e` was used later, thought just to update it here (could be updated way around, to keep here `e` and later use different variable name)

andre-merzky: *sigh*

PR #259, opened 2020-08-13
Fix to respect default values of Munch-based attributes

mtitov:
Default values of attributes of Munch-based types are ignored with method `update`

Examples are based on `Config` class (pure Munch-based) from `radical.dreamer`
```
>>> from radical.dreamer import Config
```
1) create an instance with default values:
```
>>> cfg = Config()
>>> print(pprint.pformat(cfg))
{'rabbitmq': {'exchange': 'rd',
              'queues': {'allocation': 'allocation',
                         'request': 'request',
                         'resource': 'resource',
                         'schedule': 'schedule',
                         'session': 'session',
                         'workload': 'workload'},
              'url': 'amqp://localhost:5672/'},
 'session': {'early_binding': True,
             'output_profile': 'profile.json',
             'schedule_options': []}}
```
2) instance with default values except one:
```
>>> cfg = Config({'rabbitmq': {'url': 'nourl'}})
>>> print(pprint.pformat(cfg))
{'rabbitmq': {'url': 'nourl'},
 'session': {'early_binding': True,
             'output_profile': 'profile.json',
             'schedule_options': []}}
```
If attributes are of Munch-based type with some default values, then these default values should be considered in the `update` method (thus it is only related to Munch-based attributes)

**edit:** [link](https://github.com/radical-project/radical.dreamer/blob/master/src/radical/dreamer/configs/_base.py#L34) to `Config` class

### DIFF
--- a/src/radical/utils/munch.py
+++ b/src/radical/utils/munch.py
@@ -92,19 +92,23 @@ class Munch(DictMixin):
         if not other:
             return
 
-        for k,v in other.items():
+        for k, v in other.items():
             if isinstance(v, dict):
                 t = self._schema.get(k)
                 if not t:
                     t = type(self)
                 if not isinstance(t, dict):
-                    if (issubclass(type(v), Munch)):
-                        # no need to recast)
+                    if issubclass(type(v), Munch):
+                        # no need to recast
                         pass
                     elif issubclass(t, Munch):
                         # cast to expected Munch type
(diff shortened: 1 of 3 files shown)


mtitov:
@andre-merzky yes, will do!

mtitov:
@andre-merzky could a minor release (1.5.3?) with latest changes (including this one) be published?

PR #260, opened 2020-08-17
handle defaults the same way as schema

andre-merzky:
This simplifies inheritance: the child class does not need to apply defaults on it's own, and then again apply the default dict, but both are now always applied (when set).

### DIFF
--- a/src/radical/utils/munch.py
+++ b/src/radical/utils/munch.py
@@ -64,10 +64,10 @@ class Munch(DictMixin):
         self._data = dict()
 
         if defaults:
-            self.update(defaults)
+            self.update(copy.deepcopy(defaults))
 
         elif hasattr(self, '_defaults'):
-            self.update(self._defaults)
+            self.update(copy.deepcopy(self._defaults))
 
         self.update(from_dict)
 
@@ -99,7 +99,6 @@ class Munch(DictMixin):
             return
 
         for k, v in other.items():
-            v = copy.deepcopy(v)
             if isinstance(v, dict):
                 t = self._schema.get(k)
                 if not t:
@@ -107,10 +106,7 @@ class Munch(DictMixin):
                 if isinstance(t, type) and \
                         issubclass(t, Munch) and not issubclass(type(v), Munch):
                     # cast to expected Munch type
-                    options = {'from_dict': v}
-                    if self._data.get(k):
-                        options['defaults'] = self[k]
-                    v = t(**options)
+                    v = t(from_dict=v, defaults=self._data.get(k))
             self[k] = v
 
 
@@ -136,7 +132,7 @@ class Munch(DictMixin):
         # Munch-based type then `verify` method will raise TypeError exception
         #
         # attrs `_defaults` and `_data` will be deepcopied in `update` method
-        return type(self)(from_dict=self._data)
+        return type(self)(from_dict=copy.deepcopy(self._data))
 
 
     # --------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


mtitov:
@andre-merzky can you please have a look to the latest updates here. Your proposal with input parameter `defaults` was really the right approach that helped to make a better update for method `update` (using `merge` internally wasn't a good approach)

andre-merzky:
The deepcopy implementation makes sense to me, looks good.  I'm a bit worried about `deepcopy(v)` in update: that can be an arbitrary type, and we have no idea how expensive that could be.  Yeah, I know, I usually fight premature optimization - but here I would like to avoid premature pessimization ;-)  Is the reason to introduce the deepcopy there only to cover `_schema` and `_defaults`?

src/radical/utils/munch.py
```diff
@@ -93,22 +99,19 @@ def update(self, other):
...
-            else:
-                self[k] = v
+                if isinstance(t, type) and \
+                        issubclass(t, Munch) and not issubclass(type(v), Munch):
+                    # cast to expected Munch type
+                    options = {'from_dict': v}
+                    if self._data.get(k):
+                        options['defaults'] = self[k]
+                    v = t(**options)
+            self[k] = v
```
andre-merzky, line 114: How is that different from 
```py
v = t(from_dict=v, defaults=self._data.get(k))
```
?

mtitov: yes! (I rushed with that)

mtitov:
@andre-merzky yeah, I actually like it to use for `defaults` (both input parameter and internal attribute), but not sure about `_schema` (since it shouldn't be changed) p.s. without deepcopy `test_description` was failed

andre-merzky:
> @andre-merzky yeah, I actually like it to use for `defaults` (both input parameter and internal attribute), but not sure about `_schema` (since it shouldn't be changed) p.s. without deepcopy `test_description` was failed

`defaults` is deep copied in liine 67 anyway though?

andre-merzky:
> `defaults` is deep copied in liine 67 anyway though?

Nevermind, the class member is not deep-copied.  But let me benchmark this first: now we would have a copy in each task description in RP for example, so I want to make sure this doesn't give us surprised...

mtitov:
@andre-merzky should we force inherited classes `Config` and `Description` to have `defaults` as input parameter? or should look for a work around (not to use `defaults` in `update` directly)?

src/radical/utils/munch.py
```diff
@@ -99,18 +99,14 @@ def update(self, other):
...
                 if not t:
                     t = type(self)
                 if isinstance(t, type) and \
                         issubclass(t, Munch) and not issubclass(type(v), Munch):
                     # cast to expected Munch type
-                    options = {'from_dict': v}
-                    if self._data.get(k):
-                        options['defaults'] = self[k]
-                    v = t(**options)
+                    v = t(from_dict=v, defaults=self._data.get(k))
```
mtitov, line 109: Should we remove input parameter `defaults` from `Munch` and just keep internal attribute `self._defaults` (then wouldn't need to change `Config` and `Description` classes)? And update current code as:
```python
                    self._data.setdefault(k, t()).update(v)
                    continue
```
p.s. kinda input `defaults` might be not that necessary (?)

andre-merzky: Yeah, I kind of agree with the latter proposal.  We might the also remove the input schema again, and also keep the internal `self._schema` only.  The idea was to allow specifying a Munch on the fly, dynamically, but (a) I don't have a real use case for it, and we can re-add it ones one emerges, and (b) one would need both schema and defaults for that.

mtitov:
@andre-merzky I've made corresponding updates and removed `schema` and `defaults` from input. Would that work?

andre-merzky:
Thanks @mtitov !

PR #261, opened 2020-09-23
Hotfix/demunch

mtitov:
Adding possibility to go through `list` and `tuple` attributes, additionally to `dict`, for demunching brings some overhead, and speaking about performance here is some rough estimations:
```
Timings for demunch operation
No  list                            :  0.0014410159999999894
Has list with 10   elements         :  0.007918190999999991
Has list with 100  elements         :  0.063767672
Has list with 1000 elements         :  0.651458193
Has list with 1000 ru.Munch elements:  1.660170599
```
(these numbers are generated by script `examples/demunch_timing.py`)


src/radical/utils/munch.py
```diff
@@ -183,18 +196,28 @@ def __repr__(self):
...
-                    out[k] = v
-            return out
+    @classmethod
+    def demunch(cls, src):
+        '''
+        iterate given object and apply `Munch.as_dict()` to all munch type
+        values, and return the result (effectively a shallow copy).
+        '''
+        if isinstance(src, dict):
+            tgt = dict(map(lambda x: (x, cls._demunch_value(src[x])), src))
```
andre-merzky, line 210: I am not sure I like the the (`dict(map(lambda ...`) because of readability (you may want to look up Kernighan's law :-)) and would prefer dict comprehension (below).  But then again I appreciate the symmetry of the three cases, and don't think it makes a difference performance-wise.  Performance is not great, but grows linearly with list size as expected, so that's acceptable I guess.  Thanks for adding this!

```py
tgt = {k: cls._demunch_value(v) for k,v in src,items()}
```
?

andre-merzky (approved):

src/radical/utils/munch.py
mtitov: I was about to challenge that `map(lambda...` construction should outperform other constructions :) BUT your approach is 0.1-0.2s faster (for 1000 elements in dict)! And surprisingly (for me) `list` construction of the same approach is also faster compare to `list(map(lambda...`. Oh, my world is changed :D

new numbers:
```
Timings for demunch operation
No  list                            :  0.00135648599999999
Has list with 10   elements         :  0.012719005000000005
Has list with 100  elements         :  0.06359857300000002
Has list with 1000 elements         :  0.615174667
Has list with 1000 ru.Munch elements:  1.4756722220000003
Has dict with 1000 elements         :  0.6485707490000001
```

andre-merzky: Haha, this is funny :-)  Yeah, performance is surprising in Python sometimes (in both directions, unfortunately)...

PR #262, opened 2020-10-06
Resolve ResourceWarning: unclosed file

eirrgang:
Make sure that opened files are closed.

### DIFF
--- a/src/radical/utils/get_version.py
+++ b/src/radical/utils/get_version.py
@@ -68,7 +68,8 @@ def get_version(paths=None):
         # check if there is also an SDIST near the version_path
         sdist_path = version_path.replace('/VERSION', '/SDIST')
         try:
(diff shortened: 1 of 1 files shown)


src/radical/utils/get_version.py
```diff
@@ -68,7 +68,8 @@ def get_version(paths=None):
         # check if there is also an SDIST near the version_path
         sdist_path = version_path.replace('/VERSION', '/SDIST')
         try:
-            sdist_name = open(sdist_path).read().strip()
+            with open(sdist_path) as fh:
+                sdist_name = fh.read().strip()
```
eirrgang, line 72: Shouldn't this (also) have an `rstrip()`?

andre-merzky (approved):
Thanks!  I think the `strip()` is fine, we want to be resilient against any spacing really...

PR #263, opened 2020-10-15
Added method `fcntl.lockf` in case of exception for `fcntl.flock` (file lock for counters)

mtitov:


andre-merzky (approved):
LGTM, Thanks!

PR #265, opened 2020-11-19
small plugin manager fixes

andre-merzky:
This PR cleans the `PluginManager` code.  The biggest change should be that one can now name the plugin class freely, all else are just cosmetic changes.

The PM is now used in the exaworks API implementation, thus the update.

### DIFF
--- a/src/radical/utils/plugin_manager.py
+++ b/src/radical/utils/plugin_manager.py
@@ -166,6 +166,7 @@ class PluginManager(object):
 
                 modname = os.path.splitext(os.path.basename(pfile))[0]
                 try:
+                    # load and register the plugin
                     loader = imp_loader.SourceFileLoader(modname, pfile)
                     spec   = imp.spec_from_loader(loader.name, loader)
                     plugin = imp.module_from_spec(spec)
--- a/src/radical/utils/plugin_manager.py
+++ b/src/radical/utils/plugin_manager.py
@@ -188,11 +188,11 @@ class PluginManager(object):
                     spec   = imp.spec_from_file_location(modname, pfile)
                     plugin = imp.module_from_spec(spec)
                     spec.loader.exec_module(plugin)
-#
 
                     # get plugin details from description
                     ptype  = plugin.PLUGIN_DESCRIPTION.get('type',        None)
                     pname  = plugin.PLUGIN_DESCRIPTION.get('name',        None)
+                    pclass = plugin.PLUGIN_DESCRIPTION.get('class',       None)
                     pvers  = plugin.PLUGIN_DESCRIPTION.get('version',     None)
                     pdescr = plugin.PLUGIN_DESCRIPTION.get('description', None)
 
@@ -203,6 +203,10 @@ class PluginManager(object):
 
                     if not pname:
                         self._log.error('no plugin name in %s' % pshort)
+                        continue
+
+                    if not pclass:
+                        self._log.error('no plugin class in %s' % pshort)
                         continue
 
                     if not pvers:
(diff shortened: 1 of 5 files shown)


src/radical/utils/plugin_manager.py
```diff
@@ -188,11 +188,11 @@ def _load_plugins(self):
                     spec   = imp.spec_from_file_location(modname, pfile)
                     plugin = imp.module_from_spec(spec)
                     spec.loader.exec_module(plugin)
```
mtitov, line 190: All changes look good to me! But may I suggest the following for the unchanged part (L170-190):
```python
# update import
from importlib  import util as imp, machinery as imp_loader
...
# starts at L170
                modname = os.path.splitext(os.path.basename(pfile))[0]
                try:
                    loader = imp_loader.SourceFileLoader(modname, pfile)
                    spec   = imp.spec_from_loader(loader.name, loader)
                    plugin = imp.module_from_spec(spec)
                    loader.exec_module(plugin)
```
I would call it more straightforward and it eliminates re-definition of the full module name (shown below)
```python
# L170-171
                pmodname = '%s.plugins.%s' % (self._namespace,
                                   os.path.basename(os.path.dirname(pfile)))
```

andre-merzky: it removes the `plugins` portion though?  Do the unit tests still pass with that change?

mtitov: > it removes the plugins portion though?

Andre, do you mean operations with `pmodname`? They wouldn't be needed in this case (so this operation `__import__(pmodname)` also would be eliminated). 

Tests passed: every `plugin`, which didn't have an exception, was loaded (same behavior as without these changes)

andre-merzky:
Hmm, when I plug in your lines, I get
```sh
$ ./tests/unittests/test_plugin_manager.py 
  File "./tests/unittests/test_plugin_manager.py", line 74, in <module>
    test_plugin_manager()
  File "./tests/unittests/test_plugin_manager.py", line 21, in test_plugin_manager
    plugin_2 = pmgr.load('unittests_1', 'default_2')
  File "/mnt/home/merzky/radical/radical.utils.devel/ve3/lib/python3.8/site-packages/radical/utils/plugin_manager.py", line 305, in load
    raise LookupError('No such plugin type %s in %s'
LookupError: No such plugin type unittests_1 in []
```

and that is what I expected, kind off.  This line:
```py
pmodname = '%s.plugins.%s' % (self._namespace
```

adds `.plugins` to the name, and specifies that the manager looks into a `plugins/` subdir for, well, the plugins.  The replacement code does not do that, so the plugins under `plugins/` are not found.  What am I missing?

mtitov:
@andre-merzky I just re-checked it and no errors on my side (I copied all that lines from my previous post instead of that mentioned # lines). Looks like something was missed and inside that `try...except` exceptions kept raising. Can you please check this [sub-branch](https://github.com/radical-cybertools/radical.utils/tree/fix/plugin_manager_extra)

andre-merzky:
> @andre-merzky I just re-checked it and no errors on my side (I copied all that lines from my previous post instead of that mentioned # lines). Looks like something was missed and inside that `try...except` exceptions kept raising. Can you please check this [sub-branch](https://github.com/radical-cybertools/radical.utils/tree/fix/plugin_manager_extra)

I got it now: line 137 specifies the `plugins` subdir, and it enters with the `pfile` then which is fully qualified.  Sorry that it took me a while - that i buggered up my test only confirmed what expected to find :-P

Either way: yeah, I am fine with that change, and merged this here.  Another question though: the `modname` is not unique anymore, for example if a `default` plugin is loaded for two name spaces (like, `radical.pilot.scheduler` and `radical.pilot.launcher`.  Are you sure that this does not lead to conflicts in this version (it did in the previous code)?

mtitov:
as I understand there shouldn't be any conflict, since it is just a loader and then plugin data is stored with 2 keys (`self._plugins[ptype][pname]`) BTW there are 2 test plugins with the same name: `unittests_1/plugin_unittests_default_2.py` and `unittests_2/plugin_unittests_default_2.py`

p.s. even tiny refactoring feels nice :)

mtitov (approved):

andre-merzky:
> p.s. even tiny refactoring feels nice :)

hehe, it does! :-)  Thanks!

PR #266, opened 2020-11-23
revert 9ce48d7

andre-merzky:
Sorry @mtitov , I have to revert that change.  It breaks plugins in [`radical.psi`](git@github.com:ExaWorks/prototype-saga.git) (see the `feature/radical_prototype` branch) - you can verify this by installing that package, install RS and RU, and run the single example.  With the current pmgr, relative imports in the plugin are broken and it does not load, presumably some package registration step is missing in the current scheme.

### DIFF
--- a/src/radical/utils/plugin_manager.py
+++ b/src/radical/utils/plugin_manager.py
@@ -82,9 +82,8 @@ class PluginManager(object):
 
     The plugin code consists of two parts:  a plugin description, and a plugin
     class.  The description is a module level dictionary named
-    `PLUGIN_DESCRIPTION`, the plugin class must be named `PLUGIN_CLASS`, and
-    must have a class constructor `__init__(*args, **kwargs)` to create plugin
-    instances for further use.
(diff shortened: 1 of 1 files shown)


mtitov:
@andre-merzky yeah, since plugins could load other internal modules, the name of the plugin module should be with full path, I missed that..
```python
                modname = '%s.plugins.%s.%s' % (
                    self._namespace,
                    os.path.basename(os.path.dirname(pfile)),
                    os.path.splitext(os.path.basename(pfile))[0])
```
With this version `pmodname` is not needed then: `pmodname` and `modname` could be merge into one as an example above, and `__import__(pmodname)` is not needed

p.s. in `radical.psi` there was something strange with modules loading.. RU and RS were not seen until I removed this [`__init__.py`](https://github.com/ExaWorks/prototype-saga/blob/feature/radical_prototype/radical/src/radical/__init__.py), something was wrong with namespace definition

andre-merzky:
Ok, I pushed an iteration along those lines...

mtitov (approved):

PR #267, opened 2020-11-23
Feature/env isolation

andre-merzky:
utilities to support RP's env isolation effort


andre-merzky:
ping @mtitov: this is now ready for review - thanks for your patience :-)

src/radical/utils/env.py
```diff
@@ -0,0 +1,237 @@
+
+import re
+import os
+import hashlib
+import tempfile
+
+import radical.utils as ru
```
mtitov, line 7: since this module is inside `ru`, better to use a relative import `from .misc import as_list`

src/radical/utils/env.py
```diff
@@ -0,0 +1,237 @@
...
+        # store last key/val if we have any
+        if key and key not in BLACKLIST:
+            env[key] = val
+
+    return env
+
+
+# ------------------------------------------------------------------------------
+#
+def env_prep(src, rem=None, pre=None, tgt=None):
```
mtitov, line 78: A stylistic comment: I would propose different names for input parameters, more intuitive to state the content and possible type, e.g., `kv, unset_keys, pre_exec, script_path`

src/radical/utils/env.py
```diff
@@ -0,0 +1,237 @@
...
+
+    # empty pre settings are ok - just ensure correct type
+    pre = ru.as_list(pre)
+
+    # cache lookup
+    cache_key = str(src) + str(rem) + str(pre)
+    cache_md5 = hashlib.md5(cache_key.encode('utf-8')).hexdigest()
+    if cache_md5 in _env_cache:
+
+        cache_env = _env_cache[cache_md5]
```
mtitov, line 113: should be `env = ...`

src/radical/utils/env.py
```diff
@@ -0,0 +1,237 @@
...
+        #
+        # FIXME: better tmp file names to avoid collisions
+
+        tmp_file, tmp_name = tempfile.mkstemp()
+
+        # use a file object to simplify byte conversion
+        fout = os.fdopen(tmp_file, 'w')
+        try:
+            if rem:
+                fout.write('\n# unset\n')
```
mtitov, line 141: one more stylistic comment: move first `\n` before `if rem:...`

src/radical/utils/env.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+    # if `tgt` is specified, create a script with that name which unsets the
+    # same names as in the tmp script above, and exports all vars from the
+    # resulting env from above (thus storing the *results* of the pre_exec'ed
+    # env, not the env and pre_exec directives themselves).
+    #
+    # FIXME: files could also be cached and re-used (copied or linked)
+    if tgt:
+        with open(tgt, 'w') as fout:
```
mtitov, line 185: related to this part and to the previous with tmp file, since it appears a duplication - propose - firstly create a string with all of these, and write it to the file if `tgt` is set, then for tmp file extend this string with pre-exec commands and write to that tmp file. Thus it would decrease the number of writes to the file (once per file) and eliminate the duplication

src/radical/utils/env.py
```diff
@@ -0,0 +1,237 @@
...
+                fout.write('# pre\n')
+                for cmd in pre:
+                    fout.write('%s\n' % cmd)
+                fout.write('\n')
+
+        finally:
+            fout.close()
+
+        cmd = '/bin/sh -c ". %s && env | sort > %s.env"' % (tmp_name, tmp_name)
+        os.system(cmd)
```
mtitov, line 171: should `sh_callout` (from `.shell`) be used here?

src/radical/utils/env.py
```diff
@@ -0,0 +1,237 @@
...
+
+    for k in keys_1:
+        v = env_1[k]
+        if   k not in env_2: only_1[k]  = v
+        elif v != env_2[k] : changed[k] = [v, env_2[k]]
+
+    for k in keys_2:
+        v = env_2[k]
+        if   k not in env_1: only_2[k]  = v
+        elif v != env_1[k] : changed[k] = [env_1[k], v]
```
mtitov, line 231: this line is not needed since it was already checked earlier

tests/unittests/test_env.py
```diff
@@ -0,0 +1,58 @@
...
+def test_prep_env():
+
+    try   : del(os.environ['TEST'])
+    except: pass
+
+    src = dict(os.environ)
+    ret = ru.env_prep(src)
+    only_src, only_ret, changed = ru.env_diff(src, ret)
+    assert(not only_ret), only_ret
+    assert(not changed), changed
```
mtitov, line 23: my test got failed here... looks like this `cmd = '/bin/sh -c ". %s && env | sort > %s.env"' % (tmp_name, tmp_name)` brings something extra. Here is what I've got for `only_src, only_ret, changed`: `{} {'SHLVL': '2'} {'XPC_SERVICE_NAME': ['application.com.jetbrains.pycharm.5317648.23892166', '0']}`. It is related to my local env setup, but also shows that some changes can come out, thus this part should be re-thinked

src/radical/utils/env.py
andre-merzky: thanks, done.

src/radical/utils/env.py
andre-merzky: but... then they don't align anymore???  

Just kidding (a bit) - will fix!

src/radical/utils/env.py
andre-merzky: duh

src/radical/utils/env.py
andre-merzky: done

src/radical/utils/env.py
andre-merzky: we could do that.

src/radical/utils/env.py
andre-merzky: That would not really work: the script *with* the `pre_exec`s is executed, and the *resulting* env is stored in the script file - that new env is expected to be different than the original env.  Consider the pre_exec is `['export PATH=/tmp/bin:$PATH']`, then te original `PATH` we write in the tmp script is different from the `PATH` we write in the target script.

src/radical/utils/env.py
andre-merzky: but... but... symmetry???

thanks, I removed it.

tests/unittests/test_env.py
andre-merzky: Ah, that's interesting!  `SHLVL` we should exclude, the `XPC_SERVICE_NAME` is funny...  Any idea where this is coming from?

andre-merzky: `XPC_SERVICE_NAME`: I assume that you run the test from pycharm, and pycharm sets that env for each created process, so the result is indeed different from `os.environ`.  I am not sure what to do about it, as that is cause by the test environment, not by the test or code.  Maybe just filter that env var?

andre-merzky: I pushed an iteration, can you tray again, please?

mtitov: btw here I mostly meant that it is possible that some env vars could appear in that lists, as unexpected outcomes (and in my case there were that env vars). Kinda it shouldn't depend on how we run tests

p.s. tests still got failed, but looks like `XPC_SERVICE_NAME` appears in `env = dict(os.environ)` first (when I deleted the same way you did but for `env` then it worked)

mtitov: as follow up (couldn't figure out why `XPC_SERVICE_NAME` is not set in `env` correctly)
```
# inside `env_prep`

cmd = '/bin/sh -c ". %s && echo $XPC_SERVICE_NAME && env | sort"' % tmp_name
out, err, ret = sh_callout(cmd)
print(out)

> application.com.jetbrains.pycharm.5317648.23892166
> COMMAND_MODE=unix2003
> CONDA_DEFAULT_ENV=ru-dev
> ...
> XPC_SERVICE_NAME=0
```
thus env var `XPC_SERVICE_NAME` was exported correctly (1st line in output), right? and that tmp file contains the following `export XPC_SERVICE_NAME='application.com.jetbrains.pycharm.5317648.23892166'\n`

andre-merzky: Yes, it seem that the tmp file contains the non-`0` setting and that it is set.  You can try:
```py
        cmd = '((. %s && echo "1: $XPC_SERVICE_NAME" && env && echo "2: $XPC_SERVICE_NAME") | (sort; echo "3: $XPC_SERVICE_NAME"); echo "4: $XPC_SERVICE_NAME")' % tmp_name
```
to maybe see where that env creeps in - but at the end it does not matter, really, as pycharm is not a case we necessarily need to cover.  Google is not very helpful here, as `XPC_SERVICE_NAME` is often listed in various outputs, but rarely the topic of any discussion.

Do the tests work with the last change, and do they work outside of pycharm?

mtitov: looks like XPC is a Mac-related service ([link](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingXPCServices.html)) and may be such behavior of env variable is truely exceptional (my worry was that if such case exists, that env variable is exported but then `env` returns a different value for this env var, then could it be applied to some HPC machines environment). Running tests from the terminal doesn't show error related to this env var.

Thus we can either skip this env var or add it to `BLACKLIST` only

p.s.
> cmd = '((. %s && echo "1: $XPC_SERVICE_NAME" && ...

for this test I got the following (used earlier created tmp file):
```
$ ((. /var/folders/dx/l6xvg1qd20jclmyc5cn0g5q80000gn/T/tmp_mgrt96l && echo "1: $XPC_SERVICE_NAME" && env && echo "2: $XPC_SERVICE_NAME") | (sort; echo "3: $XPC_SERVICE_NAME"); echo "4: $XPC_SERVICE_NAME")
1: application.com.jetbrains.pycharm.5317648.23892166
2: application.com.jetbrains.pycharm.5317648.23892166
COMMAND_MODE=unix2003
CONDA_DEFAULT_ENV=ru-dev
CONDA_EXE=/Users/matitov/Projects/_miniconda/bin/conda
CONDA_PREFIX=/Users/matitov/Projects/_miniconda/envs/ru-dev
...
XPC_FLAGS=0x0
XPC_SERVICE_NAME=0
_=/usr/bin/env
_CE_CONDA=
_CE_M=
__CFBundleIdentifier=com.jetbrains.pycharm
__CF_USER_TEXT_ENCODING=0x1F5:0x0:0x0
3: 0
4: 0
```

andre-merzky: > > cmd = '((. %s && echo "1: $XPC_SERVICE_NAME" && ...
> 
> for this test I got the following (used earlier created tmp file):

Ah, stupid me - the command did not let the env escape the `(..)` subshell, so the above is expected, but does not give any new insight.  Sorry...

> Thus we can either skip this env var or add it to BLACKLIST only

I'd rather handle it in the tests than in the blacklist - the code functions in the expected way, really - confirming that in the specific test environment is what we have trouble with...

mtitov: @andre-merzky can you please look at my last commit here - to clean `os.environ` first helped, since in `env.env_prep` there is another call for `os.environ`

andre-merzky: > to clean `os.environ` first helped, since in `env.env_prep` there is another call for `os.environ`

Neat! :-)

andre-merzky:
test are failing on a ZMQ timing error, which I would like to not address right now (it's a test problem, not a ZMQ problem).  Flake8 fails with the usual `imported symbol not being used` error from `__init__.py` which this PR touches.  Any suggestion how to handle that?

mtitov (approved):

PR #272, opened 2020-12-09
Added GitHub Actions to run tests

iparask:


andre-merzky (approved):
Thanks, lets see how it goes! :-)  This LGTM - apart from the failing tests.  Any idea?

mtitov:
I did fix test for locks, but for `test_metric_expand` - should be loaded module `radical.pilot` - put in dependencies for tests?

mtitov (approved):

andre-merzky:
> I did fix test for locks, but for `test_metric_expand` - should be loaded module `radical.pilot` - put in dependencies for tests?

Thanks for looking into that!  Yes, we probably should add RP as dependency.

iparask:
> > I did fix test for locks, but for `test_metric_expand` - should be loaded module `radical.pilot` - put in dependencies for tests?
> 
> Thanks for looking into that! Yes, we probably should add RP as dependency.

Why we need RP as a dependency? It is a bit counterintuitive for me. Also, I see that the locks test fail as well.

mtitov:
Test `test_metric_expands` contains the following `{'ru.STATE': 'rp.PMGR_ACTIVE'   }`, which would require to have `radical.pilot`.

p.s. it is not clear what happening with `test_locks`, test is passed locally for me, but here it looks differently

iparask:
I get the same error locally as we see in `tests`:
```python
========================================================================================================================================================== FAILURES ==========================================================================================================================================================
_________________________________________________________________________________________________________________________________________________________ test_locks _________________________________________________________________________________________________________________________________________________________

    def test_locks():
        '''
        Test debug lock wrappers
        '''

        os.environ['RADICAL_DEBUG']        = 'True'
        os.environ['RADICAL_DEBUG_HELPER'] = 'True'

        import radical.utils as ru

        l  = ru.Lock()
        rl = ru.RLock(name='bar')

        assert(not l.waits)
        assert(not rl.waits)

        with l:
            with rl:
                assert(not l.waits)
                assert(not rl.waits)

>       assert(l.name  in ru.debug._debug_helper.locks)                       # noqa
E       AttributeError: 'NoneType' object has no attribute 'locks'

l          = <radical.utils.debug.Lock object at 0x7fdbc3402e10>
rl         = <radical.utils.debug.RLock object at 0x7fdbc340a2d0>
ru         = <module 'radical.utils' from '/Users/iparask/opt/anaconda3/envs/rct_devel/lib/python3.7/site-packages/radical/utils/__init__.py'>

tests/unittests/test_locks.py:31: AttributeError
```

I'll see if I can debug this.

andre-merzky:
@iparask : can you try running it stand-alone as:
```sh
$ pytest tests/unittests/test_locks.py::test_locks
```

and then again together with multiple tests?  Stand-alone should work, collectively fail.  I think the reason is that the DebugHelper is a singleton, and is *not* created if RU is imported the first time w/o the `RADICAL_DEBUG` env being set.  Now, the specific test sets that env before import - but I assume that pytest caches imported modules, and thus the env setting has no effect.  If I add this before the import:
```py 
    import sys
    for m in sys.modules:
        if 'radical' in m:
            print(m, sys.modules[m])
```

I see the following modules cached:
```
radical <module 'radical' (namespace)>
radical.utils.atfork.atfork <module 'radical.utils.atfork.atfork' from '/mnt/home/merzky/radical/radical.utils.devel/ve3/lib/python3.8/site-packages/radical/utils/atfork/atfork.py'>
...
radical.utils.lockable <module 'radical.utils.lockable' from '/mnt/home/merzky/radical/radical.utils.devel/ve3/lib/python3.8/site-packages/radical/utils/lockable.py'>
radical.utils.singleton <module 'radical.utils.singleton' from '/mnt/home/merzky/radical/radical.utils.devel/ve3/lib/python3.8/site-packages/radical/utils/singleton.py'>
...
radical.utils.debug <module 'radical.utils.debug' from '/mnt/home/merzky/radical/radical.utils.devel/ve3/lib/python3.8/site-packages/radical/utils/debug.py'>
...
radical.utils <module 'radical.utils' from '/mnt/home/merzky/radical/radical.utils.devel/ve3/lib/python3.8/site-packages/radical/utils/__init__.py'>
```

Ergo: the env needs to be set before invoking pytest.

iparask:
@andre-merzky, I did, and it worked as you expected. The test fails when executed along with the other tests and succeeds on its own. Is what you describe the expected behavior, or is it a bug?

andre-merzky:
> expected behavior, or is it a bug

expevcted behavior I guess, when running under pytest.  So the env needs to be set before running pytest.

PR #273, opened 2020-12-11
fix munchification

andre-merzky:
This is needed by radical-cybertools/radical.pilot/pull/2287: if the type defaults to the top level type, then the initialization can recurse endlessly.  It also does not make sense to force all dict-like attributes of, say, a `PilotDescription` to be also of that type (`PilotDescription`) - it is sufficient to cast to `Munch`.  We make an exception for `ru.Config` types though, where that hierarchy makes sense.

### DIFF
--- a/src/radical/utils/config.py
+++ b/src/radical/utils/config.py
@@ -139,6 +139,8 @@ from .singleton  import Singleton
 #
 class Config(Munch):
 
+    _self_default = True
+
     # FIXME: we should cache config files after reading, so that repeated
     #        instance creations do not trigger a new (identical) round of
     #        parsing.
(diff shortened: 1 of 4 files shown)


mtitov:
@andre-merzky importing `Config` in `Munch` brings a recursion (Munch -> Config -> Munch), my tests got failed...

I see your point by changing `type(self)` to `Munch`, but would like to propose an option for this (in case we would build other classes without schemas) - to have a class attribute (flag) that would define default Munch-based class:
```python
class Munch(DictMixin):

    _self_default = False

    def update(self, other):
        ...
                if not t:
                    if self._self_default: t = type(self)
                    else                 : t = Munch
        ...
```
and then just need to add to `Config` class `_self_default = True` 

p.s. it is also possible to have this class attribute with actual class/type applied, but it could go out of control, that's why with flag is more convinient

andre-merzky:
Neat, works for me - would you mind adding that and applying to the `ru.Config` class?

mtitov:
sure, on it

mtitov (approved):

PR #274, opened 2020-12-25
Feature/get lfile owner

andre-merzky:
cleanup of lockfile class, add more tests, documentation.

### DIFF
--- a/src/radical/utils/lockfile.py
+++ b/src/radical/utils/lockfile.py
@@ -3,10 +3,11 @@
 import os
 import time
 import errno
-import fcntl
 
-from .misc  import as_bytes
-from .debug import get_caller_name
+import threading as mt
+
+from .misc   import as_bytes
+from .debug  import get_caller_name
 
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/radical/utils/lockfile.py
```diff
@@ -45,17 +57,45 @@ class Lockfile(object):
 
     # --------------------------------------------------------------------------
     #
-    def __init__(self, fname):
+    def __init__(self, fname, delete=False, *args, *8kwargs):
```
mtitov, line 60: typo next to `kwargs` (-> `**kwargs`)

src/radical/utils/lockfile.py
```diff
@@ -96,59 +144,114 @@ def acquire(self, timeout=None):
...
+        That name can then be retrieved by any other process or thread which
+        attempts to lock the same file.  Otherwise this method returns
+        'unknown'.
+        '''
+
+        if not self._delete:
+            return 'unknown'
+
+        if not os.path.isfile(self._fname):
+            return None
```
mtitov, line 178: should it return `'unknown'` as well?

p.s. and may be re-group?
```python
        output = 'unknown'
        if self._delete and os.path.isfile(self._fname):
            try:
                with open(self._fname, 'r') as fin:
                    output = fin.read()
            except:
                pass
        return output
```

src/radical/utils/lockfile.py
```diff
@@ -96,59 +144,114 @@ def acquire(self, timeout=None):
...
+
+        try:
+            with open(self._fname, 'r') as fin:
+                return(fin.read())
```
mtitov, line 182: may be better move `return` out from `with` statement?

tests/unittests/test_lockfile.py
```diff
@@ -1,9 +1,16 @@
...
 import time
+import errno
 import pytest
 
+
+import multiprocessing as mp
+import radical.utils   as ru
+
+
 import radical.utils as ru
```
mtitov, line 14: duplication

src/radical/utils/lockfile.py
andre-merzky: Oops...  Well, it did shut up the linter :-P

src/radical/utils/lockfile.py
andre-merzky: > should it return `'unknown'` as well?

When the lockfile does not exist anymore, it is not locked by anybody - so `None` is actually correct.

src/radical/utils/lockfile.py
andre-merzky: Hmm, why?  The return will leave the scope, and the `with` resource context will be released.

tests/unittests/test_lockfile.py
andre-merzky: Oh, thanks

src/radical/utils/lockfile.py
mtitov: should it be checked first then?

src/radical/utils/lockfile.py
mtitov: yeah, maybe python can handle it, just the construction doesn't look good, but ok

src/radical/utils/lockfile.py
andre-merzky: > should it be checked first then?

Fair, I changed the order.

src/radical/utils/lockfile.py
andre-merzky: Yeah, I think that's fine - it's the whole purpose of the `with` construct really, to ensure cleanup on any form of scope exit...

andre-merzky:
I have no idea why the tests are failing - do you?

mtitov:
@andre-merzky I think it might be related to this [L136](https://github.com/radical-cybertools/radical.utils/blob/b8acfd47116ee8d2f0efc50629140640b452be4d/src/radical/utils/lockfile.py#L136), seems that `fcntl.flock` doesn't return anything (== always returns `None`)

btw should we also consider same locking as for IDs -> [link](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/ids.py#L262)

andre-merzky:
> @andre-merzky I think it might be related to this [L136](https://github.com/radical-cybertools/radical.utils/blob/b8acfd47116ee8d2f0efc50629140640b452be4d/src/radical/utils/lockfile.py#L136), seems that `fcntl.flock` doesn't return anything (== always returns `None`)
> 
> btw should we also consider same locking as for IDs -> [link](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/ids.py#L262)

`lockf` is not thread safe: any thread releasing a `lockf` lock will also release the lock for all other threads holding it.  I don't mind much for IDs where we usually have a single point in code where that is handled, but I do mind for the generic lock file.  On Theta, it's probably a file system specific issue (do you happen to know on what file system type the `$RADICAL_BASE` lived?

`flock` *is* thread safe - but it is not defined by POSIX, but by BSD.  Even NFS supports it nowadays (under linux), but it's not unlikely that there are shared file systems around which don't.  

Either way, I'd rather not fall back to `lockf` in this case.  Hmmm...  Nothing is ever easy...

andre-merzky:
@mtitov : I have to ask you for a new review on this: I switched from `fcntl` to symlinks for locking...  This should be more portable now.  We probably don't want to use this for the IDs, as it will likely not scale well on shared file systems (it hits the metadata servrs a couple of times).

mtitov (approved):
wow, completely new code :) LGTM!

p.s. sorry, I wasn't clear with my last comment - I meant not to switch but to use both (with `try...except`), but it is not the case for now. And for Theta it was regarding GPFS

andre-merzky:
>  I meant not to switch but to use both (with `try...except`),

Yes, understood - but that would change the semantics of the lock to 'not thread safe', which I wanted to avoid...

PR #276, opened 2021-01-04
Feature/stack ns

andre-merzky:
generalize the stack query to support other name spaces than just `radical`.

Alas, this PR also removes the `-v` flag from `radical-stack`, as pip does not support the respective query anymore.  We need to look for an alternative way to implement this (related to radical-cybertools/radical.entk/issues/476)

### DIFF
--- a/src/radical/utils/stack.py
+++ b/src/radical/utils/stack.py
@@ -54,7 +54,7 @@ def stack(ns='radical'):
                 try:
                     ret[namespace][mname] = import_module(mname).version_detail
                 except Exception as e:
-                    if '%s_DEBUG' % ns.upper() in os.environ:
+                    if '%s_DEBUG' % namespace.upper() in os.environ:
                         ret[namespace][mname] = str(e)
                     else:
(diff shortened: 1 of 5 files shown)


src/radical/utils/stack.py
```diff
@@ -0,0 +1,66 @@
...
+                mbase = os.path.basename(mpath)
+                mname = '%s.%s' % (namespace, mbase)
+
+                if mbase.startswith('_'):
+                    continue
+
+                try:
+                    ret[namespace][mname] = import_module(mname).version_detail
+                except Exception as e:
+                    if 'RADICAL_DEBUG' in os.environ:
```
mtitov, line 57: should env var also be updated? `'RADICAL_DEBUG'` -> `'%s_DEBUG' % namespace.upper()`

andre-merzky: Thanks, makes sense.

mtitov (approved):
LGTM

andre-merzky:
Thanks, appreciated! :-)

PR #277, opened 2021-01-04
capture current env

andre-merzky:
Small change to `prep_env` to default to the environment the method is called in (instead of requiring a captured source environment)


mtitov (approved):
LGTM!

PR #279, opened 2021-01-13
Fix profiler data flushing

andre-merzky:
For some reason or the other, closing a profile does not seem to flush
remaining data to disc.  Specifically, the RP agent scheduling component
failed to flush `unschedule` events during termination, even though the
`Profiler.close()` is explicitly called and flushes.  It might be
a threading issue, as those events are issued from a different thread
than the one which created the Profiler instance.

This patch enables flush-on-write until the above problem is cleanly
solved.

### DIFF
--- a/src/radical/utils/profile.py
+++ b/src/radical/utils/profile.py
@@ -4,7 +4,7 @@ import csv
 import time
 
 from   .ids     import get_radical_base
-from   .misc    import as_string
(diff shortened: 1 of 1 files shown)


mtitov (approved):
LGTM

andre-merzky:
Thanks!

PR #280, opened 2021-01-23
make profiler fork-save

andre-merzky:
```
# The profiler is not using threads and is generally threadsafe (all write ops
# should be atomic) - but alas Python threadlocks I/O streams, and those locks
# can still deadlock after fork
#
#   - https://bugs.python.org/issue6721
#   - https://bugs.python.org/issue40399
#
# We thus have to close/reopen the prof file handle after fork.  This creates
# a bit of a mess as we not have to maintain a global list of profiler instances
# to clean up after fork... :-/
```

### DIFF
--- a/src/radical/utils/profile.py
+++ b/src/radical/utils/profile.py
@@ -3,9 +3,6 @@ import os
 import csv
 import time
 
-import threading as mt
-
-
 from .ids     import get_radical_base
 from .misc    import as_string, as_list
 from .misc    import get_env_ns      as ru_get_env_ns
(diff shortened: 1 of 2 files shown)


src/radical/utils/profile.py
```diff
@@ -99,10 +102,45 @@ def _sync_ntp():
...
+
+
+def _atfork_child():
+    global _profilers
+    for prof, fname in _profilers:
+        prof._handle = open(fname, 'a', buffering=1024)
+
+
+# lock cleaning can be disabled by setting RADICAL_UTILS_NO_ATFORK
+if 'RADICAL_UTILS_NO_ATFORK' not in os.environ:
```
mtitov, line 136: this env variable is also checked inside `atfork`, maybe we can skip it here?

andre-merzky: Yep, you are right - thanks, this is fixed now.

mtitov (approved):
LGTM

PR #281, opened 2021-01-30
improve env handling, add env eval shell quotation

andre-merzky:
scattered fixes and changes to env prep and evaluation, to support RP env isolation


andre-merzky:
The failing test is again ZMQ, otherwise tests pass...

andre-merzky:
@mtitov : ping

src/radical/utils/env.py
```diff
@@ -23,6 +23,51 @@
...
+            line = line.strip()
+
+            if not line:
+                continue
+
+            if line.startswith('#'):
+                continue
+
+            cmd, spec = line.split(' ', 1)
+            if cmd == 'unset':
```
mtitov, line 58: also to collect it into some list/set (`unset`) and in section `env` check that `k not in unset`, since `unset` could come before `export` for the same variable

src/radical/utils/shell.py
```diff
@@ -13,38 +13,43 @@
...
+def sh_callout(cmd, stdout=True, stderr=True, shell=False, env=None):
     '''
     call a shell command, return `[stdout, stderr, retval]`.
     '''
 
     # convert string into arg list if needed
-    if not shell and is_string(cmd): cmd = shlex.split(cmd)
+    if is_string(cmd) and \
+       not shell: cmd    = shlex.split(cmd)
+    if not env  : env    = os.environ
```
mtitov, line 24: this one is not actually needed, since with `None` env of the current process will be taken (same for other 2 methods)
> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment.

tests/unittests/test_env.py
```diff
@@ -53,11 +53,61 @@ def test_prep_env():
...
+
+    fname = '/tmp/env.%d' % os.getpid()
+    os.system('/bin/sh -c "env > %s"' % fname)
+    try:
+        env = ru.env_read(fname)
+
+        for k,v in env.items():
+            assert(os.environ[k] == v)
+
+        for k,v in os.environ.items():
```
mtitov, line 68: my test fails, but because of IDE (PyCharm), `os.environ` also contains env variable `OLDPWD` (and `env` doesn't contain that), should we resolve it or just skip?

src/radical/utils/env.py
andre-merzky: The method should preserve the semantics of the shell string though, so I think keeping the order is ok.

tests/unittests/test_env.py
andre-merzky: Either works for me - any preference?

andre-merzky (commented):
Sorry for letting this slip...

src/radical/utils/env.py
```diff
@@ -107,6 +153,11 @@ def env_prep(environment, unset=None, pre_exec=None, script_path=None):
     If `script_path` is given, a shell script will be created in the given
     location so that shell commands can source it and restore the specified
     environment.
+
+    Any commands given in 'cmds' will be part of the cached script, and will
+    thus *not* be executed when preparing the env, but *will* be executed
+    whenever the prepared shell script is sources.  The returned env dictionary
+    will thus *not* include the effects of those injected commands.
```
andre-merzky, line 160: exec_cmds, preserve_cmds

tests/unittests/test_env.py
mtitov: ok to skip it

src/radical/utils/shell.py
```diff
@@ -101,9 +99,6 @@ def sh_callout_async(cmd, stdin=True, stdout=True, stderr=True,
     #       python applications.
     assert(False), 'this is broken for python apps'
 
-    if not env:
-        env = os.environ
-
     # --------------------------------------------------------------------------
```
andre-merzky, line 102: thx

andre-merzky:
@mtitov : is there anything else to do on this PR?

src/radical/utils/env.py
mtitov: Andre, this one was left for review

andre-merzky: Ah, right - thanks :-)

mtitov: as we agreed, I went through the code, where this method is used, so with changing names for these two inputs, two updates in `feature/env_isolation` branch of RP (within `mpirun` LM) would be needed. Thus I would propose to go for it and make these updates :)

andre-merzky:
closing in favor of #293

PR #282, opened 2021-02-03
Feature/descr inheritance

andre-merzky:
implement schema and default inheritance for typed dictionary


src/radical/utils/munch.py
```diff
@@ -60,11 +60,19 @@ def __init__(self, from_dict=None):
             self._schema = dict()
           # raise RuntimeError('class %s has no schema defined' % self.__name__)
 
+        if hasattr(self, '_schema_extend'):
+            extend = object.__getattribute__(self, '_schema_extend')
+            self._schema.update(extend)
```
mtitov, line 65: `self._schema.update(self._schema_extend)` it will be resolved as needed, since attribute starts with `_`

src/radical/utils/munch.py
```diff
@@ -60,11 +60,19 @@ def __init__(self, from_dict=None):
...
+
         self._data = dict()
 
         if hasattr(self, '_defaults'):
             self.update(copy.deepcopy(self._defaults))
 
+        if hasattr(self, '_defaults_extend'):
+            extend = object.__getattribute__(self, '_defaults_extend')
+            self._data.update(copy.deepcopy(extend))
```
mtitov, line 74: same here, and better to use class method `update`: `self.update(copy.deepcopy(self._defaults_extend))`

src/radical/utils/munch.py
andre-merzky: Thanks, changed now.

src/radical/utils/munch.py
andre-merzky: Ack, same

andre-merzky (commented):
Thanks, both are addressed now.

mtitov (approved):
LGTM

PR #286, opened 2021-03-04
zmq server and client classes

andre-merzky:
This is an intermediary commit towards the ZMQ changes (related to MongoDB removal).  It adds a `zmq.Server` and `zmq.Client` base classes for a synchronous `REQ/REP` ZMQ sockets.  Example server and client are includes in `examples/zmq_*.py`.

Note that this PR is toward `feature/bulk_pubsub`, not devel.


mtitov:
@andre-merzky it is still in progress, right? so far I went through and have no objections, maybe one question to clarify: why response does contain `exc` and `err`? I would think they are the same, just `error` is what you set (manually) and `exception` is an error from "outside" (automatically set), could it be just `err`? (and to extend it with return code?)

andre-merzky:
The intent for `exc` is to get something one can raise, i.e., an actual typed exception with the stack reflecting the remove call context where the exception occured.  `err` would only contain an error message.  I did not yet implement the exception serialization / deserialization, that seems not as straight-forward in Python as I thought :-/

src/radical/utils/zmq/bridge.py
```diff
@@ -16,6 +17,18 @@ class Bridge(object):
...
 
+    # --------------------------------------------------------------------------
+    #
+    @staticmethod
+    def get_config(name, pwd=None):
+
+        if not pwd: pwd = '.'
+        cfg = Config(cfg=read_json('%s/%s.cfg' % (pwd, name)))
+        with open('/tmp/t', 'a') as fout:
+            fout.write('=== cfg for %s/%s: %s\n' % (pwd, name, cfg.as_dict()))
```
mtitov, line 28: this log file is just temporary?

src/radical/utils/zmq/bridge.py
```diff
@@ -25,7 +38,8 @@ def __init__(self, cfg):
         self._uid     = self._cfg.uid
         self._pwd     = self._cfg.path
         self._log     = Logger(name=self._uid, ns='radical.utils',
-                               level=self._cfg.log_lvl, path=self._pwd)
+                               level='debug', path=self._pwd)
+                             # level=self._cfg.log_lvl, path=self._pwd)
```
mtitov, line 42: is this also temporary?

src/radical/utils/zmq/queue.py
```diff
@@ -399,17 +416,19 @@ def _get_nowait(url, timeout, log, prof):  # timeout in ms
     # --------------------------------------------------------------------------
     #
     @staticmethod
-    def _listener(url, log, prof):
+    def _listener(url, qname=None):
```
mtitov, line 419: should `_listener` in `pubsub.Subscriber` be updated the same way?

src/radical/utils/zmq/bridge.py
andre-merzky: indeed - removed.

src/radical/utils/zmq/bridge.py
andre-merzky: yes, thanks!

src/radical/utils/zmq/queue.py
andre-merzky: Not necessary - the `PubSub` channels already separates message types by `topic` (that's handled on the ZMQ layer) - this PR basically introduces a similar mechanism for the `Queue`, handled on our layer.

mtitov (approved):
LGTM!

PR #289, opened 2021-03-31
flux helper

andre-merzky:


andre-merzky:
@mtitov : ping

src/radical/utils/flux.py
```diff
@@ -0,0 +1,421 @@
...
+
+
+    # --------------------------------------------------------------------------
+    #
+    def __init__(self, name : str = None) -> None:
+
+        if name: self._name = name
+        else   : self._name = 'flux_helper'
+
+        self._log         = Logger('radical.utils.%s' % self._name)
```
mtitov, line 49: is `ns=` needed here?

src/radical/utils/flux.py
```diff
@@ -0,0 +1,421 @@
...
+
+        flux_uid = generate_id('flux')
+
+        check = 'flux env; echo "OK"; while true; do echo "ok"; sleep 1; done'
+        start = 'flux start -o,-v,-S,log-filename=%s.log' % flux_uid
+        cmd   = '/bin/bash -c "echo \\\"%s\\\" | %s"' % (check, start)
+
+        penv  = os.environ
+        if env:
+            penv  = {k:v for k,v in os.environ.items()}
```
mtitov, line 100: to avoid duplications, maybe move L100 instead of L98

src/radical/utils/flux.py
```diff
@@ -0,0 +1,421 @@
...
+        def _watch_flux(flux_env : Dict[str,str],
+                        flux_term: mt.Event
+                       ) -> None:
+
+            self._log.info('starting flux watcher')
+
+            for k,v in flux_env.items():
+                os.environ[k] = v
+
+            ret = None
```
mtitov, line 157: not needed

src/radical/utils/flux.py
```diff
@@ -0,0 +1,421 @@
...
+
+        with self._lock:
+
+            flux_info = self._flux_info.get(uid)
+
+            if not flux_info:
+                try:
+                    flux_info = read_json('%s/%s.json' % (self._base, uid))
+                except Exception as e:
+                    raise LookupError('flux service id %s unknown' % uid) from e
```
mtitov, line 209: I got confused here, so there is a possibility that this file exists, but there is no corresponding record in `self._flux_info`, then we read that data and should write into `self._flux_info`? If file exists, but there is no record in `self._flux_info` then the following L211 will fail, right? (since `self._flux_info` and `self._local_state` are set at the same time) But in the first place it is not clear how file could exist without a corresponding record, unless we can create an instance of `FluxHelper` for an old config (?)

src/radical/utils/flux.py
```diff
@@ -0,0 +1,421 @@
...
+
+        with self._lock:
+
+            flux_info = self.check_service(uid)
+            assert(uid in self._local_state)
+
+            try:
+                from flux import job as flux_job
+                args = {'url':flux_info['uri']}
+                jex  = flux_job.executor.FluxExecutor(handle_kwargs=args)
```
mtitov, line 260: should we reuse already imported `flux` module? (`self._mod.job.executor.FluxExecutor(...)`)

mtitov:
@andre-merzky added some comments inline, but I didn't touch `Munch` here, keep it for other PR

src/radical/utils/flux.py
andre-merzky: Not really, it will be [derived](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/logger.py#L232) from the given name, so `ns=...` is only useful when the name does not reflect the expected namespace (`Logger('pmgr.000.log', ns='radical.pilot')`)

src/radical/utils/flux.py
andre-merzky: I changed the default to `None` - as you said somewhere else, `popen` will default to `os.environ` in that case, so we can skip this step.

src/radical/utils/flux.py
andre-merzky: ack

src/radical/utils/flux.py
andre-merzky: You are right, that was incorrect.  Also the separation of `local_state` and `flux_info` was a bit unclear in other places, I tried to clean it up.

src/radical/utils/flux.py
andre-merzky: ack, done.

src/radical/utils/flux.py
```diff
@@ -220,6 +221,12 @@ def check_service(self, uid : str) -> Dict[str, Any]:
                                           'executors': list(),
                                           'callbacks': list()}
 
+            else:
+                if self._local_state[uid]['term'].is_set():
+                    # the service is gone already
+                    raise RuntimeError('flux service id %s unknown' % uid)
```
mtitov, line 227: just a leftover (not a critical one) - since `uid` is known, error message should be about that it was terminated/stopped

mtitov (approved):
LGTM! (there is a suggestion only left)

src/radical/utils/flux.py
andre-merzky: makes sense, fixed - thanks!

PR #290, opened 2021-04-08
multiple namespaces to load plugins from

andre-merzky:
This can be used to collect plugins from different installation trees.


mtitov (approved):
LGTM!

p.s. I also fixed `release` methods for Lock(s), since they return nothing (was a pylint complain)

andre-merzky:
Thanks @mtitov !

PR #291, opened 2021-04-08
make check on set and type cast optional

andre-merzky:


src/radical/utils/munch.py
```diff
@@ -299,11 +360,33 @@ def verify(self, ctype=None):
...
+
+
+        if not self._check:
+            # no type checking on set
+            return v
+
+        if self._schema:
+
+            if k.startswith('__'):
+                return v
```
mtitov, line 381: (part 1) we have some inconsistency with attributes names if they start with `_`:
- if we read or write attribute that starts with `_` (e.g., `munch_obj._name`) then it is not written into `self._data`
- for items (e.g., `munch_obj['_name']`) all names are accepted (at least before changes)
```
>>> class TestMunch(Munch):
...     _check = True
...     _schema = {'_name': str}
...     _defaults = {'_name': 'misha'}
... 
>>> m = TestMunch()
>>> m._name
  File "<stdin>", line 1, in <module>
  File "/Users/matitov/Projects/_miniconda/envs/stack/lib/python3.8/site-packages/radical/utils/munch.py", line 160, in __getattr__
    return object.__getattribute__(self, k)
AttributeError: 'TestMunch' object has no attribute '_name'
>>> m['_name']
'misha'
```

(part 2) this `_verify_setter` doesn't allow to set item if there is no schema, but for attribute it works if it starts with `_`
```
>>> class TestMunch(Munch):
...     _check = True
... 
>>> m=TestMunch()
>>> m._test = 'hello'
>>> m._test
'hello'
>>> m['_test'] = 'bye'
>>> m['_test']
>>> dir(m)
['_test']
```

src/radical/utils/munch.py
```diff
@@ -221,56 +231,98 @@ def as_dict(self):
...
     @classmethod
-    def _verify_bool(cls, k, v, t):
-        if v is None: return
-        if v              in [True, False]       : return v
-        if str(v).lower() in ['true', 'yes', '1']: return True
-        if str(v).lower() in ['false', 'no', '0']: return False
-        raise TypeError('%s: expected bool type for %s (%s)'
-                       % (cls.__name__, k, type(v)))
+    def _verify_bool(cls, k, v, t, cast):
+        if v in [True, False]: return v
```
mtitov, line 274: missed line `if v is None: return`

src/radical/utils/munch.py
```diff
@@ -221,56 +231,98 @@ def as_dict(self):
...
+        if cast:
+            if str(v).lower() in ['true', 'yes', '1']: return True
+            if str(v).lower() in ['false', 'no', '0']: return False
+            raise TypeError('%s: expected bool type for %s (%s)'
+                           % (cls.__name__, k, type(v)))
+        else:
+            if isinstance(v, bool):
+                return v
```
mtitov, line 282: these 2 lines are not needed since it already was checked with line `if v in [True, False]: return v`

src/radical/utils/munch.py
```diff
@@ -221,56 +231,98 @@ def as_dict(self):
     # --------------------------------------------------------------------------
     #
     @classmethod
-    def _verify_int(cls, k, v, t):
+    def _verify_int(cls, k, v, t, cast):
         if v is None: return
-        try   : return int(v)
-        except: raise TypeError('%s: expected int type for %s (%s)'
-                               % (cls.__name__, k, type(v)))
+        if cast:
```
mtitov, line 236: maybe for all verifiers to make as the first line `if v is None or isinstance(v, <type>): return v` and then check `cast` parameter?

mtitov: my bad, this proposal is not for all of them, but for `int`, `float`, `str`, `bool`

src/radical/utils/munch.py
andre-merzky: Our implementation in general [does not support underscored attributes](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/munch.py#L58), so I don't think this is a problem

src/radical/utils/munch.py
andre-merzky: Thanks, fixed

src/radical/utils/munch.py
andre-merzky: Also fixed.

src/radical/utils/munch.py
andre-merzky: Fair - the `isinstance` is likely not faster than the cast, but avoiding the `try/except` is good...

src/radical/utils/munch.py
mtitov: then I think that these lines (`if k.startswith('__'): return v`) could be removed: since in case of `__setattr__` it doesn't mater, but in case of `__setitem__` exception should be raised (since it changes `self._data` directly)

src/radical/utils/munch.py
mtitov: with these updates need to remove the same check in sections when it is `not cast` (for verifiers `int`, `str`, `float`)

mtitov:
btw in method `__deepcopy__` new changes does the same what previous approach did, but increases number of calls (+1) to `update` method

```python
        return type(self)(from_dict=copy.deepcopy(self._data))
# vs
        ret = type(self)()
        ret.update(copy.deepcopy(self._data))
        return ret
```

src/radical/utils/munch.py
```diff
@@ -338,18 +352,12 @@ def _verify_kvt(cls, k, v, t, cast):
...
-
-    def verify(self, ctype=None):
-
-        if ctype and not issubclass(type(self), ctype):
-            raise TypeError('class type mismatch: %s >< %s'
-                    % (type(self), ctype))
+        if issubclass(t, Munch)   : return cls._verify_munch(k, v, t, cast)
+        if cast: return v
+        raise TypeError('%s: no verifier defined for type %s'
+                       % (cls.__name__, t))
```
mtitov, line 358: this part `raise TypeError(..)` was commented out, would it work this way? Thus raise an error if `cast` is not set and corresponding verifier is not found

mtitov:
@andre-merzky I've added verifier for Munch-based attributes, can you please have a look

src/radical/utils/munch.py
andre-merzky: ok, done.

src/radical/utils/munch.py
andre-merzky: Also done, thanks.

src/radical/utils/munch.py
andre-merzky: I may be parsing the comment wrong, but that is indeed the expected behavior.

PR #292, opened 2021-04-08
Fix/munch defaults

andre-merzky:
Allow to set `None` as `Munch` default value, indicating that the attribute is not set at all


mtitov (approved):
LGTM!

PR #293, opened 2021-04-08
Feature/env isolation

andre-merzky:
Hmm, I could have sworn I had that PR opened before :-/  Anyway, here it is, if a bit late.  This cleans some of the env isolation support scripts, adds documentation to some others, etc.  It also adds a simple FS based process synchronization routine on the shell level (for rank `pre_exec` synchronization in RP).


andre-merzky:
@mtitov : wrt. our discussion on #281, I now changed the names to `pre_exec` and `pre_exec_cached` - that preserves the `pre_exec` nomen I would like to keep in sync with RP, but also makes the semantics clearer - does that work for you?

mtitov:
@andre-merzky names `pre_exec` and `pre_exec_cached` are good fit, but looks like names for other parameters got reverted back, from what we changed them, before the latest changes we had (1) `env_prep(environment, unset=None, pre_exec=None, cmds=None, script_path=None)` (and wanted to fix names for commands), and now we have (2) `env_prep(source=None, target=None, remove=None, pre_exec=None, pre_exec_cached=None)`

andre-merzky:
> @andre-merzky names `pre_exec` and `pre_exec_cached` are good fit, but looks like names for other parameters got reverted back, from what we changed them, before the latest changes we had (1) `env_prep(environment, unset=None, pre_exec=None, cmds=None, script_path=None)` (and wanted to fix names for commands), and now we have (2) `env_prep(source=None, target=None, remove=None, pre_exec=None, pre_exec_cached=None)`

@mtitov : this should be fixed now, linter and tests pass also.  Let me know if you are ok with merging this, please.

src/radical/utils/env.py
```diff
@@ -140,16 +140,22 @@ def env_eval(fname):
...
-            if cmd == 'unset':
+            if line.startswith('unset ') :
+                _, spec = line.split(' ', 1)
                 k = spec.strip()
                 if k not in env:
                     continue
                 del(env[k])
-            elif cmd == 'export':
+
+            if line.startswith('export ') :
```
mtitov, line 150: should be `elif`? (otherwise after unset-section it can go to L155)

andre-merzky: Ah, nice catch - fixed.

mtitov (approved):
LGTM! Thanks, Andre!

PR #294, opened 2021-04-08
Feature/bulk pubsub

andre-merzky:
Apart from scattered cleanup, the main point of this PR is  to not blow up list messages into smaller individual messages, but to forward them in bulk.  Note that the Queue will still unrol lists (and we may want to introduce a request size for repacking there) - but in the pubsub case, any consumer will get all messages anyway, so there is no point in unroling...


andre-merzky:
This PR looks a bit larger as it should as it contains the `zmq.Server` changes also - I expected this to into `devel` earlier.  Sorry for that...

mtitov:
@andre-merzky there is a failed test `test_json.test_metric_expand` I cannot figure out why it started failing, and several pylint/flake8 warnings left

mtitov (approved):
LGTM!

PR #295, opened 2021-04-12
Feature/registry

andre-merzky:
Purpose: move some of RPs config management from disk to memory


andre-merzky:
TODO AM: check status

src/radical/utils/zmq/client.py
```diff
@@ -21,43 +23,54 @@
 #
 class Request(object):
 
-    def __init__(self, cmd=None, arg=None):
+    def __init__(self,
+                 cmd: str = None,
+                 arg: Optional[List[Any]] = None
+                ) -> None:
 
         self._cmd = cmd
```
mtitov, line 31: later when property `cmd` is called it is checked that `self._cmd` is not empty, should then we not allow `None` for instance creation? -> `def __init__(self, cmd: str, arg:...)`

src/radical/utils/zmq/client.py
```diff
@@ -21,43 +23,54 @@
...
         self._arg = arg
```
mtitov, line 32: is `None` checked later? or we can keep `arg` as list always? -> `self._arg = arg or []`

src/radical/utils/zmq/client.py
```diff
@@ -76,26 +89,29 @@ def str(self):
...
+    def from_dict(cls, msg: Dict[str, Any]) -> 'Response':
 
         return Response(res=msg.get('res'),
                         err=msg.get('err'),
                         exc=msg.get('exc'))
 
 
     @property
-    def res(self): return self._res
+    def res(self) -> 'Result':
```
mtitov, line 105: `Result`? should it be `str`?

src/radical/utils/zmq/client.py
```diff
@@ -76,26 +89,29 @@ def str(self):
...
         return cls.from_dict(msgpack.unpackb(msg))
 
 
     @classmethod
-    def from_dict(cls, msg):
+    def from_dict(cls, msg: Dict[str, Any]) -> 'Response':
 
         return Response(res=msg.get('res'),
                         err=msg.get('err'),
                         exc=msg.get('exc'))
```
mtitov, line 101: since both them return the instance of this class, maybe move them into init-method?
```python
def __init__(self,
             from_dict: Optional[Dict[str, Any]] = None,
             from_msg: Optional[bytearray]       = None,
             res: Optional[str]       = None,
             err: Optional[str]       = None,
             exc: Optional[List[str]] = None
            ) -> None:
```

src/radical/utils/zmq/client.py
```diff
@@ -130,13 +146,13 @@ def __init__(self, server=None, url=None):
 
 
     @property
-    def url(self):
+    def url(self) -> Optional[str]:
```
mtitov, line 149: it is assumed that `self._url` is set during init method, thus should it be `... -> str`?

src/radical/utils/zmq/registry.py
```diff
@@ -0,0 +1,133 @@
...
+            ret = this.get(elems[-1])
+            return ret
+
+        except AttributeError:
+            return None
+
+
+    # --------------------------------------------------------------------------
+    #
+    def keys(self, arg):
```
mtitov, line 77: `def keys(self):`?

src/radical/utils/zmq/client.py
andre-merzky: Ack, fixed.

src/radical/utils/zmq/client.py
andre-merzky: Ack, fixed.

src/radical/utils/zmq/client.py
andre-merzky: I'd rather keep them separate so that the user does not have to specify parameter names.

src/radical/utils/zmq/client.py
andre-merzky: ack, fixed.

src/radical/utils/zmq/registry.py
andre-merzky: indeed - fixed.

src/radical/utils/zmq/client.py
mtitov: ping on this - should we enforce to have it as a `list` attribute?

src/radical/utils/zmq/client.py
```diff
@@ -26,24 +26,21 @@ class Request(object):
     def __init__(self,
                  cmd: Optional[str]       = None,
                  arg: Optional[List[Any]] = None
-                ) -> None:
+                 ) -> None:
```
andre-merzky, line 29: Hmm, my linter complains if the closing bracket has a different indentation than the opening one ... :-/  Does the pylint check complain about this?

src/radical/utils/zmq/client.py
andre-merzky: Thanks for the ping.  I think an empty (`None`) arg should be valid.  I'll check the code to make sure that the assumption holds.

src/radical/utils/zmq/client.py
mtitov: is it pylint? my IDE highlights it if they are on the same position, as I understand in such case when it is new line within parentheses it should be aligned with first letter after the opening parentheses

p.s. personally, in this case I would prefer to move the closing parentheses into the previous line
```python
    def __init__(self,
                 cmd: Optional[str]       = None,
                 arg: Optional[List[Any]] = None) -> None:
```

mtitov:
@andre-merzky there is an issue with `zmq.Server` - if registered method (with `register_request`) doesn't have input arguments then `rep = self._success(self._cbs[cmd](arg))` will fail ([link](https://github.com/radical-cybertools/radical.utils/blob/40747d707e1fc38b3fb7eba7bbb6cf10d70a45cc/src/radical/utils/zmq/server.py#L280))

Currently it is forced that every registered method should have only one input argument, this could be relaxed to have `0` or `1` argument:
```python
                        if not arg:
                            inputs = as_list(arg)
                        else:
                            inputs = [arg]
                        rep = self._success(self._cbs[cmd](*inputs))
```
but better to relax it more to have `0..n` input arguments, thus need to update `arg` within `zmq.Request` (or on server side, but I think it is preferable to enforce it within request)

mtitov:
Currently registry server is never actually stopped/terminated - the following lines should be added at the end of method `ru.zmq.server.Server._work` ([link](https://github.com/radical-cybertools/radical.utils/blob/cdd248ee40d0a0c2e6496c55dfa06a72ff827694/src/radical/utils/zmq/server.py#L289)):
```python
        self._sock.close()
        self._ctx.term()
```

andre-merzky:
> Currently registry server is never actually stopped/terminated - the following lines should be added at the end of method `ru.zmq.server.Server._work` ([link](https://github.com/radical-cybertools/radical.utils/blob/cdd248ee40d0a0c2e6496c55dfa06a72ff827694/src/radical/utils/zmq/server.py#L289)):
> 
> ```python
>         self._sock.close()
>         self._ctx.term()
> ```

This is not really needed (nor advised really): clean closing of ZMQ sockets seems to be a bit tricky when threads are present, and it seems to be good practice to let the garbage collector take care of the context destruction (which will close all sockets for that context also).

https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Context.destroy
https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Socket.close

From the latter: `This can be called to close the socket by hand. If this is not called, the socket will automatically be closed when it is garbage collected.`

mtitov:
@andre-merzky sorry, missed to explain why this took my attention - after having several tests in RP where Registry server was started (for RM and LM initializations from Registry), that server instances were in conflict between each other: the following server start up couldn't happen because the one before wasn't stopped.

May be we can add a method which we can call [manually] whenever we want to force to stop/close a server?
```python
    def close(self):
        self._sock.close()
        self._ctx.term()
```

mtitov:
@andre-merzky the final solution I came to is that socket should be closed within the method it is created in (since this method is related to a thread). If sockets wouldn't be inside threads then it would be possible to have a separate method to close them and to terminate a corresponding context, but outside of threads context cannot close sockets (context has method `destroy` which could close sockets, but it is not threadsafe, thus sockets should be closed not by the context).

src/radical/utils/zmq/client.py
andre-merzky: Done

andre-merzky:
This PR should be ready now.  I am not sure what to do about the failing pylint check: the test clearly passes, so the complaint (value not subscriptable) is clearly invalid.  Any suggestion?

mtitov (approved):
LGTM!

PR #297, opened 2021-04-27
Changes related to GitHub actions

mtitov:
Keep in sync with RP

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a single version of Python
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: Python application
+name: Build
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/python-app.yml
```diff
@@ -10,37 +10,41 @@ on:
...
-
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 2
     - name: Set up Python 3.6
       uses: actions/setup-python@v2
       with:
-        python-version: 3.6
+        python-version: 3.6.13
```
andre-merzky, line 23: Just  curious: what is the reason for that change?

andre-merzky (approved):
LGTM

.github/workflows/python-app.yml
mtitov: no significant reason, just thought to keep the same as for RP. Should I revert back to `3.6`? (and within some active PR will do the same for RP?)

andre-merzky: I don't have a strong opinion, really - but in general think we should use what users would get by default with 3.6, which is presumable the latest subminor.  But maybe not, who knows / who cares :-P

Having the repos in sync is good, so :+1: on that.

mtitov: Same here, no strong opinion :) ok, then keep it as it is for now! Thanks!

PR #299, opened 2021-05-03
Feature/env isolation

andre-merzky:
Aaand another iteration on the env isolation :-)  It's small changes, only adds an `EnvProcess`.  Alas, this branch has `feature/registry` already merged - the diff will be much smaller once #295 is merged...


src/radical/utils/env.py
```diff
@@ -242,6 +259,8 @@ def env_prep(environment=None, unset=None, pre_exec=None,
             if unset:
                 fout.write('# unset\n')
                 for k in sorted(unset):
+                    if not k.isalnum():
+                        continue
```
mtitov, line 263: need to extend it with `"_"`, otherwise env vars as `CONDA_EXE` will be skipped here

src/radical/utils/env.py
```diff
@@ -233,7 +250,7 @@ def env_prep(environment=None, unset=None, pre_exec=None,
         # given name and fill it with `unset` and `export` statements to
         # recreate that specific environment: any shell sourcing that
         # `script_path` file thus activates the environment we just prepared.
-        tmp_file, tmp_name = tempfile.mkstemp()
+        tmp_file, tmp_name = tempfile.mkstemp(dir=os.getcwd() + '/env/')
```
mtitov, line 253: file will be created if the corresponding directory exists, we cannot guaranty that `env/` dir exists in the current dir

src/radical/utils/env.py
```diff
@@ -255,9 +274,11 @@ def env_prep(environment=None, unset=None, pre_exec=None,
             if environment:
                 fout.write('# export\n')
                 for k in sorted(environment.keys()):
-                    if k not in BLACKLIST:
-                        fout.write("export %s=%s\n"
-                                  % (k, _quote(environment[k])))
+                    if k in BLACKLIST:
+                        continue
+                    if not k.isalnum():
+                        continue
```
mtitov, line 280: same as above with `_`

src/radical/utils/env.py
```diff
@@ -233,7 +250,7 @@ def env_prep(environment=None, unset=None, pre_exec=None,
         # given name and fill it with `unset` and `export` statements to
         # recreate that specific environment: any shell sourcing that
         # `script_path` file thus activates the environment we just prepared.
-        tmp_file, tmp_name = tempfile.mkstemp()
+        tmp_file, tmp_name = tempfile.mkstemp(dir=os.getcwd() + '/env/')
 
         # use a file object to simplify byte conversion
         fout = os.fdopen(tmp_file, 'w')
```
mtitov, line 256: should we have method `f.write` to be called once? (and at line 314)

src/radical/utils/env.py
```diff
@@ -348,5 +371,106 @@ def env_diff(env_1, env_2):
...
+
+    # --------------------------------------------------------------------------
+    #
+    def get(self) -> str:
+
+        data, exc = self._data
+        if exc:
+            raise exc
+
+        return str(data)
```
mtitov, line 472: should it be converted into string?

(*) in RP/env_isolation there is `envp.put(self._init_from_scratch(lm_cfg, env_lm, env_sh))`, which saves a `dict` value into registry (and which later will be read)

src/radical/utils/env.py
```diff
@@ -242,6 +261,8 @@ def env_prep(environment=None, unset=None, pre_exec=None,
             if unset:
                 fout.write('# unset\n')
                 for k in sorted(unset):
+                    if not k.isalnum() and not su.is_snake_case(k):
```
mtitov, line 264: I would suggest not to have external packages for such "small" operations, and particularly for this case use `re` -> `if not re.match(r'^[a-zA-Z_][\w]+$', k)`. Would this work?

src/radical/utils/misc.py
```diff
@@ -943,5 +944,28 @@ def ret(*argv):
   # return ret
 
 
+# --------------------------------------------------------------------------
+#
+def host_is_local(host: str) -> bool:
```
mtitov, line 949: (1) this method should be moved to the module `host`
(2) could it be renamed in a way as `is_...`/`has_...`: `is_localhost` (kinda more boolean-checker, e.g., `is_active`, `is_set`, `has_flag`, etc.)

src/radical/utils/env.py
andre-merzky: applied - thanks

src/radical/utils/misc.py
andre-merzky: done

src/radical/utils/env.py
mtitov: I left regular-expression part only, since it actually covers all cases, but having 2 checks together might give a wrong output
```
>>> k = '1ENV'
>>> not k.isalnum() and not re_snake_case.match(k)  # should be `True` to be skipped
False
>>> not re_snake_case.match(k)
True
```
but if variable name will be a little changed, then both results are correct
```
>>> k = '1ENV_VAR'
>>> not k.isalnum() and not re_snake_case.match(k)
True
>>> not re_snake_case.match(k)
True
```

src/radical/utils/logger.py
```diff
@@ -254,6 +253,17 @@ def __init__(self, name, ns=None, path=None, targets=None, level=None,
...
 
+        if isinstance(level, int):
+            self._debug_level = max(0, level - 10)
+
+        elif level.upper().startswith('DEBUG_'):
+            self._debug_level = int(level.split('_', 1)[1])
+            level = 'DEBUG'
+
+        else:
+            self._debug_level = 0
```
mtitov, line 264: Should corresponding PR (#309) be merged first?

src/radical/utils/zmq/server.py
```diff
@@ -163,6 +163,7 @@ def start(self) -> None:
             raise RuntimeError('`start()` can be called only once')
 
         self._thread = mt.Thread(target=self._work)
+        self._thread.daemon = True
```
mtitov, line 166: just to confirm that this line is required (since we didn't set it in the corresponding PR)

mtitov (approved):
have couple comments, but overall LGTM!

PR #300, opened 2021-05-13
handling encoded string for mongodb password

lee212:
According to [this](https://docs.mongodb.com/manual/reference/connection-string/), special characters are allowed with encoding in the URI format, and it has to be parsed before returning DB information. If it is a normal string then the `unquote_plus` doesn't do anything, keep the original string.

### DIFF
--- a/src/radical/utils/misc.py
+++ b/src/radical/utils/misc.py
@@ -10,9 +10,10 @@ import tempfile
 import itertools
 import netifaces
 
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
The linter complains about
```
Unused quote_plus imported from urllib.parse
```
Otherwise this looks good and can be merged.

andre-merzky (approved):

PR #301, opened 2021-06-14
Fix/zmq error log

andre-merzky:
This change makes sure that the stack trace is dumped which causes an ZMQ listener thread to die.


mtitov (approved):
LGTM!

PR #302, opened 2021-06-14
Fix/shebang

andre-merzky:
ensure python3


mtitov (approved):
LGTM! p.s. is it enough to have these changes for executables in `bin/` dir only? or should I extend it for all files that have shebang?

andre-merzky:
> LGTM! p.s. is it enough to have these changes for executables in `bin/` dir only? or should I extend it for all files that have shebang?

It should be applied to all scripts - but we should not have any outside of `bin/`, or do we?

mtitov:
> It should be applied to all scripts - but we should not have any outside of `bin/`, or do we?

test files have it as well, but that just for manual execution (not with `pytest`), thus ok, let's not focus on others for now. This PR is ready to merge

PR #303, opened 2021-06-14
fix flux module handling

andre-merzky:
better naming for flux module, to avoid confusion with flux handle

### DIFF
--- a/src/radical/utils/flux.py
+++ b/src/radical/utils/flux.py
@@ -51,7 +51,8 @@ class FluxHelper(object):
         self._lock = mt.RLock()
 
         try:
-            self._mod = import_module('flux')
+            self._flux     = import_module('flux')
+            self._flux_job = import_module('flux.job')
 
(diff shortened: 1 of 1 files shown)


src/radical/utils/flux.py
```diff
@@ -51,7 +51,8 @@ def __init__(self, name : str = None) -> None:
         self._lock = mt.RLock()
 
         try:
-            self._mod = import_module('flux')
+            self._flux     = import_module('flux')
+            self._flux_job = import_module('flux.job')
```
mtitov, line 55: is `self._flux_job` necessary to have? I thought it's ok to have an access to `flux.job` just with `self._flux.job`

andre-merzky: The above does not work, I think that is because `flux.job` is a binary module or something (did not bother to try to understand why...)

mtitov (approved):
LGTM

PR #304, opened 2021-06-25
Fix/issue rp 2129

andre-merzky:
Scattered small enhancements which came up during RP #2129.


src/radical/utils/zmq/utils.py
```diff
@@ -133,6 +133,12 @@ def log_bulk(log, msgs, token):
 
     msgs = as_list(msgs)
 
+    unpacked = list()
+    for msg in msgs:
+        try   : msg = msgpack.unpackb(msg)
+        except: pass
+        unpacked.append(msg)
+
```
mtitov, line 141: should after `unpacked` is filled need to reassign it? (since currently it is not used) -> `msgs = unpacked`

andre-merzky: Ack, of course - thanks!

src/radical/utils/zmq/utils.py
```diff
@@ -88,7 +88,7 @@ def get_channel_url(ep_type, channel=None, url=None):
         # example:
         #   channel `foo`
         #   url     `pubsub://localhost:1234/foo`
-        channel = os.path.basename(Url(url.path))
+        channel = os.path.basename(str(Url(url.path)))
```
andre-merzky, line 91: We probably can use `url.path` (`str`) directly then?

mtitov: `Url` does `urlparse.urlparse` within `__init__`, thus some validation might happen there

p.s. btw I don't see that this method `get_channel_url` is used somewhere.. I made an update since it was highlighted (`basename` accepts `AnyStr`)

mtitov (approved):

PR #309, opened 2021-07-22
add `debug_1`, `debug_2` etc as additional log levels

andre-merzky:


andre-merzky:
@mtitov : ping

mtitov:
have a general question - could we use `verbose` (with default as `1`) to set the details level? thus it could be used in sync with any log level (but I guess it is applicable only to two of them: `info`, `debug`)

andre-merzky:
> have a general question - could we use `verbose` (with default as `1`) to set the details level? thus it could be used in sync with any log level (but I guess it is applicable only to two of them: `info`, `debug`)

If I understand correctly, you suggest to us 
```sh
$ export RADICAL_PILOT_LOG_LVL=DEBUG
$ export RADICAL_PILOT_LOG_VERBOSE=3
```
instead of
$ export RADICAL_PILOT_LOG_LVL=DEBUG_3
```
?  What would be the advantage?  And how would that relate to `INFO` which you mention?

What would be the advantage?

**Comment by mtitov** [comment]
@andre-merzky yes, that's correct
> how would that relate to `INFO` which you mention?

to have one approach for both, since for `info` we could also have details level (don't have it now, but having one approach could make it easier to extend), avoid having `info_1,2..` in future
p.s. for `warning` and `error` it is not applicable

But I understand I might really overthinking it, and to have levels is enough only for debug mode. I'm currently on tests and will add updates soon

File: src/radical/utils/logger.py [hunk 1]
```diff
@@ -254,6 +253,17 @@ def __init__(self, name, ns=None, path=None, targets=None, level=None,
         if level in [OFF, 'OFF']:
             targets = ['null']
 
+        if isinstance(level, int):
+            self._debug_level = max(0, level - 10)
```
mtitov, line 257: as a reminder - should this be modified or skipped? (since DEBUG is not set) and set debug_level for DEBUG only?

mtitov (approved):
LGTM!

PR #312, opened 2021-08-20
Expand recognized argument types for utils.misc.as_list().

### LINKED ISSUE #311: Reconcile internal use of misc.as_list() with API argument types.
`as_list()` is frequently used to normalize inputs for RCT commands that accept singular arguments for parameters that are generically multi-valued. However, it does not recognize non-list sequences, such as tuples or generators. This imposes unnecessary constraints and even possibly performance bottlenecks (such as when filtering requests in the `raptor.Master.request_cb`) or may cause unexpected behavior (if a user passes a `tuple` or `set` of states to `rp.wait()`).

The obvious exceptions to iterables that should be allowed are `str` and `bytes`. There are RCT objects that may be acceptable individually or in groups, but I believe they all satisfy `isinstance(obj, collections.abc.Mapping)`. If not, maybe there is a root RCT object they are descended from?

In addition to (or instead of) updating `misc.as_list()`, an `as_iterable()` or `as_rct_arg()` utility function might be added and used judiciously in many of the current `as_list()` call sites.

Another alternative would be to add richer checking or casting based on an optional argument asserting the expected type of the list element, but that might be hard to read or surprising for RCT developers, and it would be hard to decide whether to try to convert arguments or elements or to just reject them.

eirrgang:
Resolves #311

### DIFF
--- a/src/radical/utils/misc.py
+++ b/src/radical/utils/misc.py
@@ -1,4 +1,5 @@
 
+import collections.abc
 import os
 import sys
 import time
@@ -368,6 +369,8 @@ def as_list(data):
 
     if   data is None : return []
     elif is_list(data): return data
+    elif isinstance(data, (str, bytes, collections.abc.Mapping)): return [data]
+    elif isinstance(data, collections.abc.Iterable): return list(data)
     else              : return [data]
 
 
@@ -943,4 +946,3 @@ def script_2_func(fpath):
 
 
 # ------------------------------------------------------------------------------
-


andre-merzky:
Hi @eirrgang - we discussed this a couple of times in our group but would like to *not* merge the PR.  The `as_list` method is used in high frequency code paths in our code, and the set of additional `isinstance` calls has a measurable performance impact which we don't think we can afford.  I'm afraid the respective type casting has to happen on the application side (from RCT perspective).

PR #313, opened 2021-09-01
Update host.py

eirrgang:
There seemed to be unnecessary logic in host.py that wouldn't work quite right,
at least for nested local hostnames.

Checking whether there is a '.' in the hostname doesn't really tell anything,
so maybe I have misunderstood the intention. However, `getfqdn()` has a
reasonable return value with various fall-back logic the makes an exception
unlikely.

Note that on MacOS, `socket.gethostbyaddr(socket.gethostname())`
is likely to raise `socket.gaierror` rather than `socket.herror`.

If this function is used for more than logging, (to determine something about the
connection between RCT components), I would think that it would be more
effective to create a trial socket and inspect it.
(E.g. `socket.create_connection(...).getsockname()`)
But I think `host.get_hostip()` is used for that sort of thing.

### DIFF
--- a/src/radical/utils/host.py
+++ b/src/radical/utils/host.py
@@ -114,7 +114,7 @@ def create_hostfile(sandbox, name, hostlist, sep=' ', impaired=False):
 
     hostlist = as_list(hostlist)
     filename = '%s/%s.hosts' % (sandbox or '.', name)
-    with open(filename, 'w') as fout:
+    with open(filename, 'w', encoding='utf8') as fout:
 
         if not impaired:
             # create dict: {'host1': x, 'host2': y}
--- a/src/radical/utils/host.py
+++ b/src/radical/utils/host.py
@@ -24,12 +24,7 @@ def get_hostname():
     global _hostname                                     # pylint: disable=W0603
     if not _hostname:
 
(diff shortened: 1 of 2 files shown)


andre-merzky (approved):
LGTM - can be merged after test passes or if amended.  Thanks!

eirrgang:
If the only purpose of this function is to provide caching, the module could be further simplified with `functools`. Note, though, that `functools.cache` is not available until Python 3.9. `functools.lru_cache` has been available for some time, but has slightly more overhead (it has to expire old values). This overhead is still infinitesimal compared to the cost of `getfqdn()`.

I've been using the following placeholder in module imports for code that will some day require Py >=3.9.

```python
try:
    cache = functools.cache
except AttributeError:
    # Note: functools.cache does not appear until Python 3.9
    cache = functools.lru_cache(maxsize=None)
```

andre-merzky:
I don't think there is anything gained by using functools, so let's leave it as it is.

eirrgang:
> I don't think there is anything gained by using functools, so let's leave it as it is.

It would eliminate the need for the unit test and remove the module global variable, but I'm fine either way.

PR #319, opened 2021-11-19
generic deactivate

andre-merzky:
This is needed for RP in cases where the pilot startup occurs into an existing VE which the bootstrapper needs to escape from.

### DIFF
--- a/bin/radical-utils-env.sh
+++ b/bin/radical-utils-env.sh
@@ -353,15 +353,10 @@ env_deactivate(){
     conda=$(which conda 2>/dev/null)
     if ! test -z "$conda"
     then
-        if test -z "$CONDA_SHLVL"
-        then
+        while test "$CONDA_SHLVL" -gt "0"
+        do
             conda deactivate
-        else
-            while ! "$CONDA_SHLVL" = "0"
-            do
-                conda deactivate
-            done
-        fi
+        done
     fi
 
     # old style conda
@@ -407,6 +402,10 @@ env_deactivate(){
         export PATH=$(echo $PATH | tr ":" "\n" | grep -v "$REMOVE" | tr "\n" ":")
         unset  $prefix
     done
+
+    # clean out CONDA env vars
+    unset  CONDA_DEFAULT_ENV
+    unset  CONDA_PROMPT_MODIFIER
 }
 
 
--- a/bin/radical-utils-env.sh
+++ b/bin/radical-utils-env.sh
@@ -344,5 +344,77 @@ sync_n(){
 
 
 # ------------------------------------------------------------------------------
+#
+env_deactivate(){
+    # provide a generic deactivate which attempts to move the current shell out
+    # of any activated virtualenv or conda env
+
+    # deactivate active conda
+    conda=$(which conda 2>/dev/null)
+    if ! test -z "$conda"
+    then
+        conda deactivate  # leave any conda env
+        conda deactivate  # leave conda base env
+    fi
+
+    # old style conda
+    has_deactivate=$(which deactivate 2>/dev/null)
+    if ! test -z "$has_deactivate"
+    then
+        . deactivate
+    fi
+
+    # deactivate active virtualenv
+    has_deactivate=$(set | grep -e '^deactivate\s*()')
+    if ! test -z "$has_deactivate"
+    then
+        deactivate
+    fi
+
+    # as a fallback, check if `$VIRTUAL_ENV` is set and dactivate manually
+    if ! test -z "$VIRTUAL_ENV"
+    then
+        # remove the `$PATH` entry if it exists
+        REMOVE="$VIRTUAL_ENV/bin"
+        PATH=$(echo $PATH | tr ":" "\n" | grep -v "$REMOVE" | tr "\n" ":")
+        REMOVE="$VIRTUAL_ENV/bin/"
+        PATH=$(echo $PATH | tr ":" "\n" | grep -v "$REMOVE" | tr "\n" ":")
+        export PATH
+        unset  VIRTUAL_ENV
+    fi
+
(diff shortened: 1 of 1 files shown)


bin/radical-utils-env.sh
```diff
@@ -343,6 +343,78 @@ sync_n(){
...
+        export PATH
+        unset  CONDA_PREFIX_1
+        unset  CONDA_SHLVL
+        unset  CONDA_EXE
+        unset  CONDA_PREFIX_1
+        unset  CONDA_PYTHON_EXE
+        unset  CONDA_PROMPT_MODIFIER
+        unset  CONDA_DEFAULT_ENV
+        unset  _CE_CONDA
+    fi
```
mtitov, line 415: no sure about whole this part..
1) `CONDA_PREFIX_1` is one from a group of such variables: in case of many activations there will be variables like `CONDA_PREFIX_x` (see example above), they are not "active", but point to previous active conda environment;
2) `PATH` doesn't contain `CONDA_PREFIX_x` values, it contains only active one (`CONDA_PREFIX`) and path to `<conda_base_dir>/condabin`
3) After deactivation, only the following variables are left (and most of them are set): `CONDA_EXE`, `CONDA_PYTHON_EXE`, `CONDA_SHLVL`, `_CE_CONDA`, and they are "harmless". And I will discourage to unset `CONDA_EXE` -> deactivation process is about to turn off virtual environment, and not to wipe out everything related to it, thus not having `CONDA_EXE` makes it impossible to re-activate conda environment without re-sourcing conda setup. I mean using `source <conda_base_dir>/etc/profile.d/conda.sh` will set necessary environment for conda (which should be executed once during bootstrapping), which will not break paths for other apps, and whenever we need conda environment we can activate (and then deactivate) it.

andre-merzky:
Great feedback, thank you, will iterate.

bin/radical-utils-env.sh
andre-merzky: Thanks, that makes sense.  I reworked this - but left the removal of the `PATH` in place.  In the best case you are right and it's not part of the path anyway and then this is a noop - in the worst case it is because of whatever and we definitely want to remove it then.

bin/radical-utils-env.sh
```diff
@@ -343,6 +343,73 @@ sync_n(){
...
+
+    # deactivate active conda
+    conda=$(which conda 2>/dev/null)
+    if ! test -z "$conda"
+    then
+        if test -z "$CONDA_SHLVL"
+        then
+            conda deactivate
+        else
+            while ! "$CONDA_SHLVL" = "0"
```
mtitov, line 360: shouldn't it be within brackets? 
btw another nice approach is to compare with `0`: `[ "$CONDA_SHLVL" -gt 0 ]`

bin/radical-utils-env.sh
mtitov: That still looks unnecessary, `conda` keeps `$CONDA_PREFIX` in `$PATH` only (no previous prefixes), other case it is if user will mess with `PATH` (and in this case we could do what user will not expect)

bin/radical-utils-env.sh
```diff
@@ -343,6 +343,78 @@ sync_n(){
...
+    # conda env fallback
+    if ! test -z "$CONDA_PREFIX"
+    then
+        # remove the `$PATH` entry if it exists
+        REMOVE="$CONDA_PREFIX/bin"
+        PATH=$(echo $PATH | tr ":" "\n" | grep -v "$REMOVE" | tr "\n" ":")
+        REMOVE="$CONDA_PREFIX/bin/"
+        PATH=$(echo $PATH | tr ":" "\n" | grep -v "$REMOVE" | tr "\n" ":")
+        export PATH
+        unset  CONDA_PREFIX
```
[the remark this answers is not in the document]
mtitov: Should we `unset` these two variables as well?

bin/radical-utils-env.sh
```diff
@@ -343,6 +343,73 @@ sync_n(){
...
+    # provide a generic deactivate which attempts to move the current shell out
+    # of any activated virtualenv or conda env
+
+    # deactivate active conda
+    conda=$(which conda 2>/dev/null)
+    if ! test -z "$conda"
+    then
+        if test -z "$CONDA_SHLVL"
+        then
+            conda deactivate
```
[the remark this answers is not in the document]
andre-merzky: Fair - done.

bin/radical-utils-env.sh
andre-merzky: Oops, forgot the `test`... - fixed

bin/radical-utils-env.sh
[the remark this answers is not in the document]
andre-merzky: I moved them out of the `if` - but yes, done now.

bin/radical-utils-env.sh
```diff
@@ -343,6 +343,72 @@ sync_n(){
...
+    do
+        # remove the `$PATH` entry if it exists
+        REMOVE="$prefix/bin"
+        export PATH=$(echo $PATH | tr ":" "\n" | grep -v "$REMOVE" | tr "\n" ":")
+        unset  $prefix
+    done
+
+    # clean out CONDA env vars
+    unset  CONDA_DEFAULT_ENV
+    unset  CONDA_PROMPT_MODIFIER
```
mtitov, line 408: ok keep it here, but better within conda related (e.g., it is not related to `VIRTUAL_ENV` case), that's why thought to have it within part if `CONDA_PREFIX` happen to be set. But ok to skip it

andre-merzky: It's a `noop` if the var's are not set, so this should be fine.

mtitov (approved):
LGTM

PR #322, opened 2021-11-21
pylint fixes, wrap open for utf8 encoding

andre-merzky:
the latter reacts on a recent pylint change

### DIFF
--- a/src/radical/utils/debug.py
+++ b/src/radical/utils/debug.py
@@ -14,6 +14,7 @@ import importlib.util
 import threading as mt
 
 from .ids     import generate_id
+from .misc    import ru_open
 from .threads import get_thread_name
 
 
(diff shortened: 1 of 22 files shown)


src/radical/utils/debug.py
```diff
@@ -258,9 +259,6 @@ def raise_on(tag, log=None, msg=None):
     process-local, but is shared amongst threads of that process.
     '''
 
-    global _raise_on_state                               # pylint: disable=W0603
-    global _raise_on_lock                                # pylint: disable=W0603
```
mtitov, line 262: don't understand why these statements are removed (and many others), since these variables are defined outside of the method, but are updated inside of it.. seems I miss something

andre-merzky: Apparently, the `global` statement is needed on
```py
foo = list()
def tmp():
   global foo
   foo = [1]
```

but not on either of these:
```py
foo = list()
def tmp():
   global foo
   foo.append(1)
   bar(foo)
```

pylint complained, I complied... :-D

andre-merzky:
Thanks for adding the test!

src/radical/utils/debug.py
mtitov: ah... because with list it does read the variable and then applies operations to its value (which is of type `list`), and not just assign a "new" value. Thanks!!

mtitov (approved):
LGTM!

PR #325, opened 2022-01-23
`TypedDict` as `Munch`-successor

mtitov:


mtitov:
Proposal to remove obsolete tests:
- `test_munch.py`
- `test_description.py`

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+
+import copy
+import sys
+
+from .misc import as_list, as_tuple, is_string
+
+
+# ------------------------------------------------------------------------------
+#
+class TDError(Exception):
```
andre-merzky, line 28: Why do we want a separate exception base class?

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+        # guaranteed class "base" attributes
+        _base_namespace = {
+            '_schema'      : {},
+            '_defaults'    : {},
+            '_self_default': False,  # convert unschemed-dict into class itself
+            '_check'       : False,  # attribute type checking on set
+            '_cast'        : True    # attempt to cast on type mismatch
+        }
+
+        for _cls in bases:
```
andre-merzky, line 64: I wonder if we get any order dependencies in the for loops below?

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+                  items
+                  iterkeys
+                  itervalues
+                  keys
+                  pop
+                  popitem
+                  setdefault
+                  update
+                  values
+                  verify
```
andre-merzky, line 116: We should check this list - for one, `merge` and `write` are currently missing.     
Do we need to support all of them?

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+        `foo` cannot be used as `from_dict` parameter to the `BarTypedDict`
+        constructor.
+        """
+        if not other:
+            return
+
+        for k, v in other.items():
+            # NOTE: if an attribute is of TypedDict and is needed not replace
+            #       it, but rather to update it, then uncomment the next 2 lines
+            # if isinstance(v, TypedDict):
```
andre-merzky, line 154: Since the method is `update`, I would assume we would always *update*, not replace.  Or am I reading this wrong?

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+            yield k, self[k]
+
+    # --------------------------------------------------------------------------
+    #
+    def get(self, key, default=None):
+        if key in self:
+            return self[key]
+        return default
+
+    def setdefault(self, key, default=None):
```
andre-merzky, line 222: Do we need this as exposed method?

andre-merzky: Same for `pop` and `popitem`

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+    # obsolete method name
+    @classmethod
+    def demunch(cls, src):
+        return cls.to_dict(src)
+
+    # --------------------------------------------------------------------------
+    #
+    @classmethod
+    def __raise_attr_error(cls, k, v, t, level=4):
+        raise TDError(
```
andre-merzky, line 316: I would expect that we would raise a proper `AttributeError` - why don't we?

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+            level=level)
+
+    @classmethod
+    def _verify_base(cls, k, v, t):
+        if cls._cast:
+            try:
+                return t(v)
+            except (TypeError, ValueError):
+                pass
+        cls.__raise_attr_error(k, v, t)
```
andre-merzky, line 327: I don't understand the semantics - if `cls._cast` is not set, we always raise.  Why is that, the type could already be correct?

andre-merzky: Same for other methods below.

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+                return False
+        cls.__raise_attr_error(k, v, t)
+
+    @classmethod
+    def _verify_tuple(cls, k, v, t):
+        if cls._cast:
+            v = as_tuple(v)
+            return tuple(
+                [cls._verify_kvt(k + ' tuple element', _v, t[0]) for _v in v])
+        else:
```
andre-merzky, line 344: Here OTOH we explicitly check type if casting is disabled.

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+            # check base types
+            if isinstance(v, t)        : return v
+            if t in [str, int, float]  : return cls._verify_base(k, v, t)
+            if t is bool               : return cls._verify_bool(k, v, t)
+            cls.__raise_attr_error(k, v, t, level=3)
+        if isinstance(t, tuple)        : return cls._verify_tuple(k, v, t)
+        if isinstance(t, list)         : return cls._verify_list(k, v, t)
+        if isinstance(t, dict)         : return cls._verify_dict(k, v, t)
+        if cls._cast                   : return v
+        raise TDError('no verifier defined for type %s' % t, level=2)
```
andre-merzky, line 397: we could try an `isinstance(v,t)` though

src/radical/utils/typeddict.py
```diff
@@ -0,0 +1,481 @@
...
+    #
+    def _verify(self):
+        """
+        Can be overloaded
+        """
+        pass
+
+    # --------------------------------------------------------------------------
+    #
+    def query(self, key, default=None, last_key=True):
```
andre-merzky, line 435: I wonder if we should overload `get()` with this functionality.  Get would first need to check if the key exists verbatim (even if the keyname has dots), but could then fallback to the recursion.
But also, we should check how frequently we use the `query` call in our code.  My hunch is that it is rare and we may not need it really and instead may want to free up that reserved word...

andre-merzky (commented):
Thanks a lot for the work on this, @mtitov !  I added a couple of questions, will also follow up on some topics myself.

src/radical/utils/__init__.py
```diff
@@ -48,6 +44,14 @@
 from .env            import env_prep, env_diff, EnvProcess
 from .stack          import stack
 
+from .dict_mixin     import DictMixin, dict_merge, dict_stringexpand, dict_diff
+from .dict_mixin     import PRESERVE, OVERWRITE, iter_diff
+from .typeddict      import TypedDict, TypedDictMeta, TDError, to_dict
+from .config         import Config, DefaultConfig
+# for backward compatibility keep the names of obsolete classes
+from .typeddict      import TypedDict as Description
```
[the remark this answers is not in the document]
mtitov: yeah, absolutely agree, I guess it is just about sync these processes (keep it here until RP/RE will be updated accordingly?)

src/radical/utils/config.py
```diff
@@ -390,34 +397,62 @@ def __init__(self, module=None, category=None, name=None, cfg=None,
...
+        if expand:
+            # NOTE: expansion is done on reference, not copy
+            ru_expand_env(src, env=env)
+
+        dict_merge(self, src, policy=policy, log=log)
+
+
+# ------------------------------------------------------------------------------
+#
+class DefaultConfigMeta(TypedDictMeta, Singleton):
```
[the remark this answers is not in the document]
mtitov: Since class `DefaultConfig` uses metaclass `Singleton`, had to "merge" it with metaclass `TypedDictMeta` from the `TypedDict` to avoid conflicts between metaclasses (kinda extended `Singleton` with `TypedDictMeta`)

src/radical/utils/config.py
```diff
@@ -390,34 +397,62 @@ def __init__(self, module=None, category=None, name=None, cfg=None,
 
         super().__init__(from_dict=cfg_dict)
 
+    # --------------------------------------------------------------------------
```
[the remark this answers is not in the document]
mtitov: Technically they are not new :) they are/were part of `Munch` class (thus part of `Config` class), but they are not used in `TypedDict` (since `TypedDict` is more about data format and not how to handle it).
The idea is to keep it for configuration class, since it is part of its functionality.

> would it make sense to remove them or maybe move them to free functions - what do you think?

I lean more towards keeping them as part of `Config` class, b/c `Config` is under our control and we know what data will be stored in config files/objects.

> (*) maybe we could also move merge to the existing update method?

not sure with this, since the behavior could be confusing (even if it is just in `Config`)

src/radical/utils/typeddict.py
mtitov: this was added just to reference the source of error (better control error message)

src/radical/utils/typeddict.py
mtitov: it is the same behavior as when we have several parent classes - `class Child(Parent_1, Parent_2)` - inherited initialization goes to `Parent_1` first and then to `Parent_2`.

src/radical/utils/typeddict.py
mtitov: we agreed not to have them in `TypedDict`, but they could be part of dedicated classes (e.g., `Config`). Most of that methods are same what `dict` type provides, but with extra ones to deal with attributes type (e.g., `verify`)

I would think they are less of general purpose, and more to specialized use cases

src/radical/utils/typeddict.py
mtitov: yes and no, I would say it depends on how we want to treat `TypedDict` instances
here is an example with dictionary:
```
>>> a= {'b': {1:1, 2:2}}
>>> a.update({'b': {3:3}})
>>> a
{'b': {3: 3}}
```
and we do the same for `TypedDict` (it is the same for `Munch`), but we can extend it by activating these commented out lines, thus existing `TypedDict` instance will be updated with data from the provided value.

My initial idea was to keep it as it is, but also to have a "new" approach for later.

src/radical/utils/typeddict.py
mtitov: Yes, to have the same methods as `dict` does

src/radical/utils/typeddict.py
mtitov: With this we have one place to set error message (not duplicate it) and `level` helps to control which method to refer here, thus not to expose private methods.

src/radical/utils/typeddict.py
mtitov: If type is already correct it will not reach this method (or other of a certain type), check L389 (only exception for `TypedDict`)

p.s. this functionality is the same as `Munch` has

src/radical/utils/typeddict.py
mtitov: probably need to give different names for these methods
- some methods are applied if `t` is a type, thus if `cast` is set then we try to convert `v`, otherwise - exception
- other methods are applied if `t` is an instance of a certain type (e.g., `[str]`), thus if `cast` is set we try to convert `v` and check its elements, if not then we check if `v` is of the same type as `t`, otherwise - exception

p.s. Andre, that is your initial idea ;) (and all that checks are valid) I only reorganized the code a little bit

src/radical/utils/typeddict.py
mtitov: yes, it is checked in L389

src/radical/utils/typeddict.py
mtitov: - I would not overload `get`, it will bring an unexpected behavior
- even if it is used rarely, I would still propose to keep it, it is nice approach for cases having many levels of dictionaries (kinda it is not specific to **typed** dict, but to its **munch** part)

src/radical/utils/typeddict.py
mtitov: I can add different flavors of errors, should I?
(I would like to keep the control part, but we can inherit from `AttributeError`, `TypeError` and `ValueError`)

mtitov: updated exceptions

src/radical/utils/__init__.py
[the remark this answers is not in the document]
andre-merzky: Works for me.

src/radical/utils/config.py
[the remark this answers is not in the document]
andre-merzky: I checked - we actually don't use `query` anywhere.  Lets comment it out, so that `cfg.query` becomes available as config attribute?
`merge` we use in exactly one place in RP (`src/radical/pilot/utils/component.py`) - and that can be covered via the `dict_merge` method I would think.

src/radical/utils/config.py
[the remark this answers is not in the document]
andre-merzky: Hmm, not sure I understand that...  What happens if `DefaultConfig` uses the previous signature?  But maybe also I don't appreciate why `TypedDictMeta` exists in the first place...

src/radical/utils/typeddict.py
andre-merzky: > we agreed not to have them in TypedDict,
Ah, of course...

src/radical/utils/typeddict.py
andre-merzky: Ack, thanks.

src/radical/utils/typeddict.py
andre-merzky: We don't use them though...  BUt I get your point...

src/radical/utils/typeddict.py
andre-merzky: Thanks, got it now.

src/radical/utils/typeddict.py
andre-merzky: Nah, names are fine - it's just too long since I looked through that code, so it took a while... Sorry for the noise...

src/radical/utils/typeddict.py
andre-merzky: See earlier comment - we only use it once.  I would opt for removing it.  We could keep it as private method (`def _query(...)`

andre-merzky (commented):
Closed most of the threads, thanks for your replies.  A few got answers :-)

src/radical/utils/typeddict.py
mtitov: ok, ack

src/radical/utils/config.py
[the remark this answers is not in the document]
mtitov: > What happens if `DefaultConfig` uses the previous signature?

with original `class DefaultConfig(Config, metaclass=Singleton):`, while `Config` is a sub-class of `TypedDict`
```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) 
subclass of the metaclasses of all its bases
```

> But maybe also I don't appreciate why `TypedDictMeta` exists in the first place...

but why? it is the only way to inherit class attributes, it is done once during class definition (dynamically),
seems nice [and right] way to prepare a class

src/radical/utils/config.py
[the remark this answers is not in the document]
mtitov: With this I don't have a strong opinion but lean towards to have these methods within `Config` class (seems like functionality that config instance would rely one)

p.s. created PR removing usage of `merge` [#2523](https://github.com/radical-cybertools/radical.pilot/pull/2523)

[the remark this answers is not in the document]
andre-merzky: As always, thanks for indulging me! :-)

src/radical/utils/config.py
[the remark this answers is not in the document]
andre-merzky: Yeah, probably right.  Small request so: can you add short docstrings to the meta classes explaining this?  Knowing me, a year from now I will wonder the same :-P

src/radical/utils/typeddict.py
andre-merzky: :+1:

andre-merzky (approved):
Thanks a lot, @mtitov - LGTM!

src/radical/utils/config.py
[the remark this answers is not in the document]
mtitov: ack

src/radical/utils/config.py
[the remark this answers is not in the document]
mtitov: updated, and just to confirm - methods `query` (renamed into `_query`) and `merge` are not in `Config` and `TypedDict`, but method `write` is part of `Config`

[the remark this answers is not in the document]
andre-merzky: Well, naturally I would like to get rid of `write()` also :-P  But yeah, the current state looks good to me.  I merged this now and will pop out the release - wohoo! :-D  Let's see what it breaks :-P

PR #333, opened 2022-02-21
add `sh_quote()`

andre-merzky:
This supports radical-cybertools/radical.pilot/pull/2522.


mtitov:
All looks good, except the test setup - I would propose to have test classes based on `TestCase`

p.s. also would propose (and start to clean that out) not to have `if __name__ == '__main__':` inside tests, which breaks the purpose of using `pytest` (e.g., specific usage of such `TestCase` methods as `setUp`, `setUpClass`, etc.)

For local runs the following command could be used `PYTHONPATH=src pytest tests/unittests/<test_module>`, instead of `python3 tests/unittests/<test_module>`

andre-merzky:
I changed the test structure to use the `TestCase` base.

mtitov (approved):
Thanks! LGTM!

PR #341, opened 2022-02-25
remove Munch and demunch from exported names

andre-merzky:
remove Munch from RU namespace to force stack consistency.

### DIFF
--- a/src/radical/utils/typeddict.py
+++ b/src/radical/utils/typeddict.py
@@ -297,11 +297,6 @@ class TypedDict(dict, metaclass=TypedDictMeta):
     def as_dict(self):
         return as_dict(self._data)
 
-    # obsolete method name
-    @classmethod
-    def demunch(cls, src):
-        return as_dict(src)
-
 
     # --------------------------------------------------------------------------
     #
@@ -488,10 +483,6 @@ def as_dict(src):
     else:
         tgt = src
     return tgt
-
-
-# keep old name(s) for previously set calls
-demunch = TypedDict.as_dict
 
 
 # ------------------------------------------------------------------------------
--- a/src/radical/utils/typeddict.py
+++ b/src/radical/utils/typeddict.py
@@ -43,6 +43,7 @@ class TDErrorMixin:
                                       f.f_code.co_name,
                                       msg and ' - %s' % msg or ''))
 
+
 # ------------------------------------------------------------------------------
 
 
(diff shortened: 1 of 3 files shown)


mtitov:
@andre-merzky should class `Description` be removed as well?

src/radical/utils/typeddict.py
```diff
@@ -107,7 +108,7 @@ class TypedDict(dict, metaclass=TypedDictMeta):
     # --------------------------------------------------------------------------
     #
     def __init__(self, from_dict=None):
-        """
+        '''
```
mtitov, line 111: hmm, as I remember we agreed to have `"""`

p.s. and within the class it is 1 line between methods

andre-merzky: Really?  But we do single quotes and double lines throughout the code base otherwise...

andre-merzky:
> @andre-merzky should class `Description` be removed as well?

Yeah, I considered that also - would be open to that for simplicity.  What do you think?
  
> **update** and remove method `demunch` that was for backward compatibility?

`demunch`: I checked the codes we have and should have catched all places where those where used (famous last words :-P - but tests should confirm).

`update`: not sure I got that: `update` is not touched by this PR?

mtitov:
> > @andre-merzky should class `Description` be removed as well?

> Yeah, I considered that also - would be open to that for simplicity.  What do you think?

yes, I'm also up for removing it (with this it will be much clear in RP)
and up to remove `demunch` from `TypedDict` (which points to `as_dict` -> `demunch = as_dict`)

> not sure I got that: update is not touched by this PR?

sorry :) this one was as "update" I made to the post, it is not about method `update`

andre-merzky:
> yes, I'm also up for removing it (with this it will be much clear in RP) and up to remove `demunch` from `TypedDict` (which points to `as_dict` -> `demunch = as_dict`)
> 
> > not sure I got that: update is not touched by this PR?

Good, will do so then!
 
> sorry :) this one was as "update" I made to the post, it is not about method `update`

Ah, got it :-D

andre-merzky:
Tests fail because the prof test pulls RP devel which does not yet handle the missing `Munch`.  That should settle once the respective PR lands in RP devel.  But that PR (radical-cybertools/radical.pilot/pull/2528) waits for this one to get merged to get `as_dict` back.  So we are running in circles a bit.  I would propose to merge this RU PR first once you approve and let the changes trickle up.

andre-merzky:
> BTW looks like tests here and in RP will fail, due to crossref...

Ah, our messages crossed, realized the same ... :)

PR #344, opened 2022-03-07
clean up temp setup files

andre-merzky:


setup.py
```diff
@@ -15,13 +16,15 @@
...
-from setuptools import setup, find_namespace_packages
+from setuptools import setup, Command, find_namespace_packages
 
 
 # ------------------------------------------------------------------------------
 name     = 'radical.utils'
 mod_root = 'src/radical/utils/'
 
+sdist_level = int(os.environ.get('SDIST_LEVEL', 0))
+os.environ['SDIST_LEVEL'] = str(sdist_level + 1)
```
mtitov, line 27: Andre, can you please give a short description of this purpose? This env variable is set by us only? Its value is changed, but this is only within this process (thus for other setup procedures it will be the same as here)?

setup.py
```diff
@@ -234,15 +246,22 @@ def read(*rnames):
...
 
-os.system('rm -rf src/%s.egg-info' % name)
+
+# ------------------------------------------------------------------------------
+# clean temporary files from source tree
+if sdist_level == 0:
+    os.system('rm -vrf src/%s.egg-info' % name)
+    os.system('rm -vf  %s/%s'           % (path, sdist_name))
+    os.system('rm -vf  %s/VERSION'      % path)
+    os.system('rm -vf  %s/SDIST'        % path)
```
mtitov, line 264: if this level is 0 then file `VERSION` will be deleted, thus we will not be able to check version of any radical package - why? (btw somehow `SDIST` wasn't deleted)

setup.py
```diff
@@ -104,59 +105,70 @@ def get_version(_mod_root):
...
 
+
+# ------------------------------------------------------------------------------
+#
+def read(fname):
     try:
-        return open(os.path.join(os.path.dirname(__file__), *rnames)).read()
+        return open(fname, encoding='utf-8').read()
     except Exception:
         return ''
```
mtitov, line 160: do we really need this function (one place where it is used is commented out)? if yes, should we update it with `with open(..) as fd` or close file explicitly?

setup.py
```diff
@@ -5,6 +5,7 @@
 __copyright__ = 'Copyright 2013-20, The RADICAL-Cybertools Team'
```
mtitov, line 5: `2013-2022` - keep it up to date? (brought it up here, just not to spread comments across other PRs - update there accordingly)

setup.py
```diff
@@ -104,59 +105,70 @@ def get_version(_mod_root):
...
-          # formerly derived version as ./VERSION
-            shutil.move('VERSION', 'VERSION.bak')              # backup version
-            shutil.copy('%s/VERSION' % path, 'VERSION')        # full version
-            os.system  ('python3 setup.py sdist')               # build sdist
+            # pip install stage 2 or easy_install stage 1
+            #
+            # pip install will untar the sdist in a tmp tree.  In that tmp
+            # tree, we won't be able to derive git version tags -- so we pack
+            # the formerly derived version as ./VERSION
+            shutil.move("VERSION", "VERSION.bak")              # backup
```
mtitov, line 127: I guess it is better to update other packages accordingly: `"` -> `'`

setup.py
andre-merzky: Yeah, I should have done that of course - doc is added now.

setup.py
andre-merzky: same doc as above applies.

setup.py
andre-merzky: Yeah, may be a good opportunity to clean it out, will do.

setup.py
andre-merzky: ack

setup.py
andre-merzky: ack

mtitov (approved):
LGTM! (took awhile to figure some details, but all looks good)

andre-merzky:
> LGTM! (took awhile to figure some details, but all looks good)

Thanks! :-)

PR #345, opened 2022-04-03
Remove RG (`radical.gtod`) from the dependencies

mtitov:
RG is required for RP (see RP PR [#2571](https://github.com/radical-cybertools/radical.pilot/pull/2571))


andre-merzky (approved):
Thanks :-)

PR #346, opened 2022-04-07
move some methods from RP.utils

andre-merzky:
see comments on radical-cybertools/radical.pilot/pull/2469/files#r839143301


mtitov (approved):
LGTM

PR #347, opened 2022-05-04
Fix I/O (error related to "too many open files")

mtitov:


andre-merzky (approved):
LStrangeTM, but oh well...

src/radical/utils/json_io.py
```diff
@@ -28,11 +28,12 @@ def read_json(fname, filter_comments=True):
     with ru_open(fname) as f:
 
         try:
-            return parse_json(f.read(), filter_comments)
-
+            output = parse_json(f.read(), filter_comments)
```
eirrgang, line 31: Ah! Good catch!

PR #348, opened 2022-07-28
fix queue pull wait time

andre-merzky:
This needs a hotfix release: the 10s queue wait time can cause unreasonably delays on queues when they are draining or underfilled.


mtitov (approved):
LGTM!

PR #351, opened 2022-08-28
fix double subscription on Subscribers which hold global zmq sockets 

andre-merzky:
fix double subscription on Subscribers which hold global zmq sockets for the same target URL

We kept `zmq.SUB` sockets in global storage to minimize the number of sockets in play.  However, that leads to double-subscription of said sockets when they are reused on the same channel, but for different topics.  This PR changes socket ownership to instance-local, thus removing that problem.


src/radical/utils/zmq/pubsub.py
```diff
@@ -283,21 +282,13 @@ def _get_nowait(socket, lock, timeout, log, prof):
     # --------------------------------------------------------------------------
     #
     @staticmethod
-    def _listener(url, log, prof):
-
-      # assert(url in Subscriber._callbacks)
+    def _listener(sock, url, log, prof, lock, term, callbacks):
```
mtitov, line 285: as some cleanup - we don't need `url` as input parameter here, since socket is provided explicitly

p.s. and as a side note - should input parameters be reordered? (e.g., as a supplementary parameters, I would keep `log` and `prof` as the end of list)

src/radical/utils/zmq/pubsub.py
```diff
@@ -327,15 +318,21 @@ def __init__(self, channel, url=None, topic=None, cb=None,
...
-        self._topics   = as_list(topic)
-        self._cb       = cb
-        self._log      = log
-        self._prof     = prof
+        Subscriber._instances.append(self)
 
-        self._uid      = generate_id('%s.sub.%s' % (self._channel,
-                                                   '%(counter)04d'), ID_CUSTOM)
+        self._channel   = channel
+        self._url       = as_string(url)
```
mtitov, line 324: general question - do we need to have these attributes? we use `channel` and `url` just to create `socket` (unless it is for re-subscription, but we don't have it now), so it is used in `__init__` only.

mtitov: same for `self._cb` - we later just can use method `subscribe` to add `topic` and `cb` into `self._callbacks`

src/radical/utils/zmq/pubsub.py
andre-merzky: ack, done.  Also added a change to make callback invocation errors non-fatal.

src/radical/utils/zmq/pubsub.py
andre-merzky: ack on the cb and that's fixed now. The url though missed the getters which I added.

mtitov (approved):
LGTM!

PR #352, opened 2022-08-29
make bridge silent by default

andre-merzky:
write no config file, logfile or profile.


andre-merzky:
TODO AM: make sure the agent needs to use `write_config`,

andre-merzky:
> TODO AM: make sure the agent needs to use `write_config`,

Confirm that RP works with these changes in place.

PR #353, opened 2022-08-30
Feature/persistent registry

andre-merzky:
As the name says: the PR adds optional persistence to the ZMQ registry.


src/radical/utils/zmq/registry.py
```diff
@@ -11,16 +13,23 @@
...
+    The `ru.zmq.Registry` is a ZMQ service which provides a hirarchical
+    persistent data store.
+    '''
 
     # --------------------------------------------------------------------------
     #
     def __init__(self, url=None,
-                       uid=None) -> None:
+                       uid=None,
+                       persistent='False') -> None:
```
mtitov, line 25: should be boolean type, right? -> `persistent=False`

src/radical/utils/zmq/registry.py
```diff
@@ -11,16 +13,23 @@
 # ------------------------------------------------------------------------------
 #
 class Registry(Server):
-
+    '''
+    The `ru.zmq.Registry` is a ZMQ service which provides a hirarchical
```
mtitov, line 17: typo `hirarchical` -> `hierarchical`

andre-merzky: Thanks!

src/radical/utils/zmq/registry.py
andre-merzky: Thanks, fixed (and type hints added)

PR #356, opened 2022-08-30
Fix/lockfile

andre-merzky:
small changes to the lockfile class - this is not used anywhere right now, but the changes were lingering for quite a while and thre is no point of keeping this branch around...


AymenFJA (approved):
LGTM.

PR #357, opened 2022-08-30
Fix/scheduler

andre-merzky:
Ensure that the `ratio` parameter to the `bisect` algorithm is correctly and consistently used.


tests/unittests/test_algorithms.py
```diff
@@ -1,157 +1,208 @@
...
+        # ----------------------------------------------------------------------
+        def schedule(n):
+            self.call_count += 1
+            return False
+        # ----------------------------------------------------------------------
+        #
+        good, bad, fail = ru.lazy_bisect(list(), schedule)
+        assert not good
+        assert not bad
+        assert not fail
```
mtitov, line 76: this part repeats code from test `test_lazy_bisect` and can be removed

tests/unittests/test_algorithms.py
```diff
@@ -1,157 +1,208 @@
...
+        for task in good:
+            assert task not in bad, (task, bad)
+            assert schedule(task) is True, task
 
-    for task in bad:
-        assert task not in good, (task, good)
-        assert schedule(task) is False, task
+        for task in bad:
+            assert task not in good, (task, good)
+            assert schedule(task) is False, task
```
mtitov, line 90: this part also repeats code from `test_lazy_bisect` and can be removed as well. Then tests are successful

PR #361, opened 2022-09-28
Feature/fast ids

andre-merzky:
Support for fast ID generation which provides process-local unique IDs w/o file access or system callouts

### DIFF
--- a/src/radical/utils/ids.py
+++ b/src/radical/utils/ids.py
@@ -19,7 +19,7 @@ TEMPLATE_SIMPLE  = "%(prefix)s.%(counter)04d"
 TEMPLATE_UNIQUE  = "%(prefix)s.%(date)s.%(time)s.%(pid)06d.%(counter)04d"
 TEMPLATE_PRIVATE = "%(prefix)s.%(host)s.%(user)s.%(days)06d.%(day_counter)04d"
 TEMPLATE_UUID    = "%(prefix)s.%(uuid)s"
-TEMPLATE_FAST    = "%(prefix)s.%(fast)s"
+TEMPLATE_FAST    = "%(prefix)s.%(fast)06d"
 
 
 _cache = {'dir'       : list(),
@@ -217,6 +217,8 @@ def generate_id(prefix, mode=ID_SIMPLE, ns=None):
 #
 def _generate_fast_id(prefix, ns=None):
 
+    print('generate fast', _fast)
+
     if not ns:
         ns = 'default'
 
@@ -228,10 +230,11 @@ def _generate_fast_id(prefix, ns=None):
     else:
         counter = _fast[ns][prefix] + 1
 
+    _fast[ns][prefix] = counter
+
     info = {'prefix': prefix,
             'fast'  : counter}
 
-    _fast[ns][prefix] = counter
 
     return TEMPLATE_FAST % info
 
@@ -338,6 +341,10 @@ def _generate_id(template, prefix, ns=None):
 #
 def reset_id_counters(prefix=None, reset_all_others=False):
 
+    global _fast
+    _fast = dict()
+
+
     if not isinstance(prefix, list):
         prefix = [prefix]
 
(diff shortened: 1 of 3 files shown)


mtitov:
@andre-merzky isn't it the same as `SIMPLE`? btw, it makes sense to update current approaches as `simple` -> `mem_counter` / `counter` (and it could be enhanced with fast-approach) and `item_counter` -> `file_counter`

andre-merzky:
> @andre-merzky isn't it the same as `SIMPLE`? btw, it makes sense to update current approaches as `simple` -> `mem_counter` / `counter` (and it could be enhanced with fast-approach) and `item_counter` -> `file_counter`

I also wanted to avoid, most of all, the code paths we have in the 'proper' ID generation.  `FAST` is 4 times faster than `SIMPLE` mainly for avoiding those.  Adding a `mem_counter` and then using a single code path gives the same semantics, but not the same performance I'm afraid.

I stumbled over this when creating millions of tasks in a test.  `SIMPLE` performs ok (about 200k / second), but `FAST` moved that to about 900k/sec - that happened to remove a bottleneck for that test I did :-)

mtitov:
Then I would propose to change the current approach:
- have special cases, e.g., `ID_UUID`, `ID_UNIQUE`, `ID_PRIVATE`, `ID_BASE`, with their specific methods (having a template for a name, but kinda not a template-approach)
- have a general case - `ID_CUSTOM` - with all that cases/options (aka template-approach - checking possible options)

Proposed `ID_FAST` is the same as `ID_SIMPLE`, but without lock mechanism (do we need that?) - maybe we can have just one of them as `ID_BASE`. Just it seems that bringing up `ID_FAST` this way breaks the consistency of the module

p.s., as I see, `ID_SIMPLE` is used in SAGA only (within RADICAL packages)

andre-merzky:
So instead of one special case we have 5 special cases?  I don't think that improves things, really (and I also consider that refactoring out of scope for this PR).  Hmmm...

mtitov:
> So instead of one special case we have 5 special cases? I don't think that improves things, really (and I also consider that refactoring out of scope for this PR).

Sorry, I disagree that it is out of scope.. while you introduce a new feature, you either follow the module design/approach, or you change it, so the new feature would fit (considering to continue support all existing cases). It seems that `ID_FAST` suppose to replace `ID_SIMPLE` (since why do we need the same functionality with two implementations?), but also the code is developed in a different way and I think it is reasonable to extend approach for a special case to other such cases (the general case - `custom` - suppose to reuse template parts from that special cases).

p.s. `ID_FASTEST` will be another new mode? ;)

(*) Have couple remarks to the current changes (will comment below), but my proposal is to rework this PR (as I wrote above)

src/radical/utils/ids.py
```diff
@@ -19,6 +19,7 @@
 TEMPLATE_UNIQUE  = "%(prefix)s.%(date)s.%(time)s.%(pid)06d.%(counter)04d"
 TEMPLATE_PRIVATE = "%(prefix)s.%(host)s.%(user)s.%(days)06d.%(day_counter)04d"
 TEMPLATE_UUID    = "%(prefix)s.%(uuid)s"
+TEMPLATE_FAST    = "%(prefix)s.%(fast)06d"
```
mtitov, line 22: should `TEMPLATE_SIMPLE` be reused? and update that template with `06d`?

andre-merzky:
I don't think it's worth the effort.  Was just response to a perf test I did, so not important enough....

tests/unittests/test_ids.py
```diff
@@ -55,14 +55,16 @@ def tearDownClass(cls):
...
     # --------------------------------------------------------------------------
     #
     def tearDown(self):
         ru.reset_id_counters(prefix=None, reset_all_others=True)
 
+
     # --------------------------------------------------------------------------
     #
-    def test_generate_id(self):
+    def test_generate_id(self, second=False):
```
mtitov, line 67: `second=False` - I assume that was used for local tests, and this is not needed here, right?

PR #363, opened 2022-11-06
Feature/get debug level

andre-merzky:
Expose the debug level via the RU logger API.


src/radical/utils/logger.py
```diff
@@ -185,7 +185,7 @@ class Logger(object):
     # --------------------------------------------------------------------------
     #
     def __init__(self, name, ns=None, path=None, targets=None, level=None,
-                 verbose=False):
+                 debug=None, verbose=False):
```
mtitov, line 188: not clear why do we need this - corresponding debug level is set with parameter `level`, which could be either string (e.g., `DEBUG_3`) or integer (e.g., 13, and then `self._debug_level` equal 3)

src/radical/utils/logger.py
```diff
@@ -264,7 +265,7 @@ def __init__(self, name, ns=None, path=None, targets=None, level=None,
         except:
             pass
 
-        self._debug_level = 0
+        self._debug_level = debug or 0
         if isinstance(level, int):
```
mtitov, line 269: should be updated what we set as a debug level - see L283-L289 - when we set `level` it means that types of logs higher than this number will be published, e.g., `level=30` then `WARNING`, `ERROR` and `CRITICAL` logs will be published (levels >= 30), thus if we set `level=10` then `DEBUG`, `INFO`, etc. logs will be published - `DEBUG` level range is between `10` and `20`

- `level=10` - all DEBUG logs (`debug_level=0`)
- `level=19` - the smallest number of DEBUG logs (`debug_level=9`)

or we can count `debug_level` from up to bottom: `level=19` -> `debug_level=1`... `level=10` -> `debug_level=10`

src/radical/utils/logger.py
andre-merzky: I found it cleaner to expose the debug level instead of having magic level names, and also makes setting and inspection symmetric (we inspect `debug_level`, and now can also set `debug_level`).

tests/unittests/test_logger.py
```diff
@@ -38,6 +38,7 @@ def tearDownClass(cls):
     #
     def test_init(self):
 
+        self._base_dir = '/tmp'
```
mtitov, line 41: this one is not needed, since we set `os.environ['RADICAL_BASE'] = '/tmp'` and `ru.get_radical_base` uses it (also it shouldn't be changed here since it is set for the whole test class)

tests/unittests/test_logger.py
```diff
@@ -76,8 +77,11 @@ def test_init(self):
     #
     def test_log_levels(self):
 
+        self._base_dir = '/tmp'
```
mtitov, line 80: same

tests/unittests/test_logger.py
```diff
@@ -76,8 +77,11 @@ def test_init(self):
...
         log_path = os.path.join(self._base_dir, 'log_lvl.log')
         l = ru.Logger('log_lvl', targets=[log_path])
+        print(l)
+        print(dir(l))
```
mtitov, line 84: debug leftovers

mtitov (commented):
have just test related comments

tests/unittests/test_logger.py
andre-merzky: Ah, right, that's a debug leftover - when running the test under a debugger (w/o pytest), the `_base_dir` is not initialized.  Will remove...

mtitov (approved):
Thanks, Andre! LGTM

PR #365, opened 2022-12-02
minor doc and setup update

andre-merzky:
When working on `setup.py`, RU usually is the blueprint for all other RCT modules.  This PR is a response to comments in those other modules and will trickle down to those.


mtitov (approved):
LGTM
p.s. and a ping about `LICENSE.md` (about copyright years there)

PR #366, opened 2022-12-13
Feature/no rmq

andre-merzky:
This accompanies https://github.com/radical-cybertools/radical.entk/pull/626.  The main purpose is to make sure that the ZMQ bridge embedded in RE stores logfiles in the session sandbox.


mtitov (approved):
LGTM

PR #367, opened 2022-12-14
fixed error related to default value None

mtitov:
Related to RP PR [#2797](https://github.com/radical-cybertools/radical.pilot/issues/2797)


andre-merzky (approved):
Thanks, LGTM

PR #369, opened 2023-01-02
fix and simplify registry path traversal

andre-merzky:
The important change is on the `put()` method where the line `this = this[elem]` needed to be de-indented.  The original code failed when adding keys to an existing entry - the keys were placed in root instead.

### DIFF
--- a/src/radical/utils/zmq/registry.py
+++ b/src/radical/utils/zmq/registry.py
@@ -75,8 +75,6 @@ class Registry(Server):
     #
     def get(self, key: str) -> Optional[str]:
 
-        self._log.debug('==== get %s', key)
-
         this  = self._data
         elems = key.split('.')
         path  = elems[:-1]
--- a/src/radical/utils/zmq/registry.py
+++ b/src/radical/utils/zmq/registry.py
@@ -55,15 +55,17 @@ class Registry(Server):
 
         this  = self._data
         elems = key.split('.')
+        path  = elems[:-1]
+        leaf  = elems[-1]
 
-        for elem in elems[:-1]:
-            step = this.get(elem)
-            if step is None:
-                step = dict()
-                this[elem] = step
-                this = this[elem]
+        for elem in path:
 
-        this[elems[-1]] = val
+            if elem not in this:
+                this[elem] = dict()
+
+            this = this[elem]
+
+        this[leaf] = val
 
         if not isinstance(self._data, dict):
             self._data.sync()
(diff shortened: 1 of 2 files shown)


src/radical/utils/zmq/registry.py
```diff
@@ -73,20 +75,20 @@ def put(self, key: str, val: Any) -> None:
...
-                if step is None:
-                    return None
-                this = step
-
-            ret = this.get(elems[-1])
-            return ret
-
-        except AttributeError:
-            return None
+        self._log.debug('==== get %s', key)
```
mtitov, line 78: should we remove this or add logging to other methods as well? just to keep consistent

tests/unittests/test_zmq_registry.py
```diff
@@ -20,6 +20,12 @@ def test_zmq_registry(mocked_prof):
         assert r.addr
         c = ru.zmq.RegistryClient(url=r.addr)
 
+        c.put('foo.bar.buz', {'biz': 11})
+        assert c.get('foo') == {'bar': {'buz': {'biz': 11}}}
+        assert c.get('foo.bar.buz.biz') == 11
+        assert c.get('foo.bar.buz.biz.boz') is None
+        assert c.get('foo') == {'bar': {'buz': {'biz': 11}}}
+
```
mtitov, line 28: no sure we need all these asserts since they repeat the following ones, but having `c.put('foo.bar.buz', {'biz': 11})` looks good (aka creating the key)

mtitov (commented):
have a general wondering - should we consider something as the following case (speaking about reassignment and changing the type of a value if needed)
```python
c = ru.zmq.RegistryClient(url=r.addr)

c.put('foo.bar.buz', 11)
c.put('foo.bar.buz.biz', 42)
```
which will raise an exception
```
.....
               this[leaf] = val
           TypeError: 'int' object does not support item assignment
```

but the following will pass as expected
```python
c.put('foo.bar.buz', 11)
c.put('foo.bar.buz', {'biz': 42})
```

p.s. overall - nice and good catch!)

src/radical/utils/zmq/registry.py
andre-merzky: Ah, the log can be removed I think, that's a leftover.  I don't think we need constant logging, and it's easily added back when need for debugging arises...

tests/unittests/test_zmq_registry.py
andre-merzky: right, I removed some test repetition.

andre-merzky:
> have a general wondering - should we consider something as the following case (speaking about reassignment and changing the type of a value if needed)
> 
> ```python
> c = ru.zmq.RegistryClient(url=r.addr)
> 
> c.put('foo.bar.buz', 11)
> c.put('foo.bar.buz.biz', 42)
> ```
> 
> which will raise an exception
> 
> ```
> .....
>                this[leaf] = val
>            TypeError: 'int' object does not support item assignment
> ```
> 
> but the following will pass as expected
> 
> ```python
> c.put('foo.bar.buz', 11)
> c.put('foo.bar.buz', {'biz': 42})
> ```

You are right, but I think the behavior is consistent: if a leaf is treated as a path element, an exception is appropriate.  I don't think there is a good way to guess the user intention.

In the second case, the path is explicitly pruned and replaced by a dict, so that's all right in my book.  The basically same behavior is what you would get on a standard dictionary:

```py
#!/usr/bin/env python3

c = {'foo': {'bar': {'buz': None}}}

# this works
c['foo']['bar']['buz'] = 11
c['foo']['bar']['buz'] = {'biz': 42}

# this raises
c['foo']['bar']['buz'] = 11
c['foo']['bar']['buz']['biz'] = 42
```

```sh
$ python3 t.py
  File "t.py", line 12, in <module>
    c['foo']['bar']['buz']['biz'] = 42
TypeError: 'int' object does not support item assignment
```

mtitov (approved):
Thanks, Andre, and yeah, I agree, just wanted some confirmation/clarification. LGTM!

PR #382, opened 2023-04-22
remove invalid strip

andre-merzky:
The removed `strip` sometimes removed significant whitespace (e.g., for VE promp variables).

### DIFF
--- a/src/radical/utils/env.py
+++ b/src/radical/utils/env.py
@@ -152,7 +152,7 @@ def env_read_lines(lines: List[str]) -> Dict[str, str]:
                 env[key] = val
 
             key = this_key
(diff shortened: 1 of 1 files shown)


mtitov:
@andre-merzky there are also strips in `env_eval` (for values), that should be updated as well, right?

andre-merzky:
> @andre-merzky there are also strips in `env_eval` (for values), that should be updated as well, right?

I think those are fine actually - those spaces should not be significant.  At least they did not trip the tests in the way the fixed one did :-)

mtitov (approved):
LGTM!

AymenFJA (approved):

PR #391, opened 2023-10-06
added ifaddress to prioritize net interface for Frontier

mtitov:
Resolves https://github.com/radical-cybertools/radical.pilot/issues/3053


andre-merzky (approved):
Perfect!

PR #392, opened 2023-10-07
avoid file collissions

andre-merzky:
This is related to RP issue 3055 - the PR attempts to avoid similar file name collisions.


src/radical/utils/config.py
```diff
@@ -28,7 +28,7 @@
     module         = 'module'
     module_path    = radical.utils.debug.find_module(module)
     usr_base_dir   = os.environ.get('RADICAL_CONFIG_USER_DIR') or \
-                     os.environ.get('HOME', '/tmp')
+                     os.environ.get('HOME', '/tmp/radical.%(user_id)')
```
mtitov, line 31: don't think we should change this - here we pick a directory where `.radical` sub-directory might be located

mtitov (approved):
@andre-merzky if you agree with changes, it is ready for merging

andre-merzky:
@mtitov - thanks for completing this!

PR #393, opened 2023-10-24
prefer infiniband over ethernet

andre-merzky:
This patch will let RU prefer )the first) infiniband interface over other (ethernet) interfaces.

### DIFF
--- a/src/radical/utils/host.py
+++ b/src/radical/utils/host.py
@@ -55,6 +55,7 @@ def get_hostip(req=None, log=None):
     req = as_list(req)
 
     white_list = [
+        'ib0',      # infiniband (e.g., Amarel)
         'hsn0',     # Frontier (HPE Cray EX)
         'ipogif0',  # Cray's
         'br0',      # SuperMIC


mtitov (approved):
LGTM!

AymenFJA (approved):

andre-merzky:
Sorry guys, Amarel tricked me (or, in other words, I was stupid).  Turns out that infiniband is not accessible from the compute nodes either, had to switch to a different interface after all.

src/radical/utils/host.py
```diff
@@ -55,6 +55,8 @@ def get_hostip(req=None, log=None):
     req = as_list(req)
 
     white_list = [
+        'ens1f1',   # amarel
+        'ib0',      # infiniband
```
mtitov, line 59: should we then not to add `ib0`?

andre-merzky: I am a bit torn.  It seems IB should be used if available - but then again, that did not work on Amarel :-P  I commented it out now.

mtitov: this works!) thanks!

mtitov (approved):

PR #395, opened 2023-11-13
Disable reporter by default

### LINKED ISSUE #394: `REPORTER` should be disabled by default
This has come up a couple of times, most recently [here](https://github.com/Parsl/parsl/pull/2923#pullrequestreview-1726886265). Note this then requires to *enable* the reporter in RP examples.

andre-merzky:
This fixes #394

### DIFF
--- a/src/radical/utils/configs/utils_default.json
+++ b/src/radical/utils/configs/utils_default.json
@@ -3,7 +3,7 @@
     "log_lvl"    : "${RADICAL_DEFAULT_LOG_LVL:ERROR}",
     "log_tgt"    : "${RADICAL_DEFAULT_LOG_TGT:.}",
     "log_dir"    : "${RADICAL_DEFAULT_LOG_DIR:$PWD}",
-    "report"     : "${RADICAL_DEFAULT_REPORT:TRUE}",
+    "report"     : "${RADICAL_DEFAULT_REPORT:FALSE}",
     "report_tgt" : "${RADICAL_DEFAULT_REPORT_TGT:stderr}",
     "report_dir" : "${RADICAL_DEFAULT_REPORT_DIR:$PWD}",
     "profile"    : "${RADICAL_DEFAULT_PROFILE:TRUE}",
--- a/src/radical/utils/configs/utils_default.json
+++ b/src/radical/utils/configs/utils_default.json
@@ -3,7 +3,7 @@
     "log_lvl"    : "${RADICAL_DEFAULT_LOG_LVL:ERROR}",
     "log_tgt"    : "${RADICAL_DEFAULT_LOG_TGT:.}",
     "log_dir"    : "${RADICAL_DEFAULT_LOG_DIR:$PWD}",
-    "report"     : "${RADICAL_DEFAULT_REPORT:True}",
+    "report"     : "${RADICAL_DEFAULT_REPORT:TRUE}",
     "report_tgt" : "${RADICAL_DEFAULT_REPORT_TGT:stderr}",
     "report_dir" : "${RADICAL_DEFAULT_REPORT_DIR:$PWD}",
     "profile"    : "${RADICAL_DEFAULT_PROFILE:TRUE}",
--- a/src/radical/utils/configs/utils_default.json
+++ b/src/radical/utils/configs/utils_default.json
@@ -3,7 +3,7 @@
     "log_lvl"    : "${RADICAL_DEFAULT_LOG_LVL:ERROR}",
     "log_tgt"    : "${RADICAL_DEFAULT_LOG_TGT:.}",
     "log_dir"    : "${RADICAL_DEFAULT_LOG_DIR:$PWD}",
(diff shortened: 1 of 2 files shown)


AymenFJA (approved):
LGTM thanks @andre-merzky.

mtitov:
With the second thought, I think `Reporter` should be enabled by default, and inside RP we set it disabled by default then.

In the RP examples when we create reporter **outside** of RP - it is unexpected to have an argument to enable it explicitly. But having reporter inside RP it is on us to integrate it properly, thus inside `Session` we can set `self._reporter = ru.Reporter(..., enabled=False)`. I personally don't care from where that default value comes (either from default config file, or from such input argument), but it should be in sync with the corresponding env variable (`RADICAL_REPORT`), which should be able overwrite it.

```python
def __init__(self, name, ns=None, path=None, targets=None, enabled=True):

    env_report = ru_get_env_ns('report', ns)
    if env_report is not None:
        env_report = str(env_report).lower()

        if env_report in ['0', 'false', 'off']:
            self._enabled = False
        else:
            self._enabled = True
    else:
        self._enabled = bool(enabled)
```

andre-merzky:
> With the second thought, I think `Reporter` should be enabled by default, and inside RP we set it disabled by default then.
> 
> In the RP examples when we create reporter **outside** of RP - it is unexpected to have an argument to enable it explicitly. But having reporter inside RP it is on us to integrate it properly, thus inside `Session` we can set `self._reporter = ru.Reporter(..., enabled=False)`. I personally don't care from where that default value comes (either from default config file, or from such input argument), but it should be in sync with the corresponding env variable (`RADICAL_REPORT`), which should be able overwrite it.
> 
> ```python
> def __init__(self, name, ns=None, path=None, targets=None, enabled=True):
> 
>     env_report = ru_get_env_ns('report', ns)
>     if env_report is not None:
>         env_report = str(env_report).lower()
> 
>         if env_report in ['0', 'false', 'off']:
>             self._enabled = False
>         else:
>             self._enabled = True
>     else:
>         self._enabled = bool(enabled)
> ```

The problem we are trying to address though is that the reporter is enabled by default.  RP (and RU) are libraries, and they should not write anything to stdout / stderr by default, unless something unexpected occurs.  Changing the default to `True` makes the PR somewhat useless in the context of the original issue I am afraid.

mtitov:
> I don't think that an env variable should overwrite an explicit argument

my proposal was about considering that that input argument is a default value (`enabled=` -> `default=`) instead of taking it from a configuration file.

> So we would need to pass an argument flag to the (primary) session.

sorry, this one I didn't understand. I thought if to have such input argument, then we would need to update Reporter creation in just this method https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/session.py#L1073-L1083 with `default=False`. And it would be aligned with - if we want to have a certain default behavior for Reporter in RP, then it should be changed there.

p.s. anyway I don't want to push it further, just it looks less intuitive for user (i.e., creating an instance of `Reporter` which is not active unless we provide an extra argument)

src/radical/utils/configs/utils_default.json
```diff
@@ -3,7 +3,7 @@
     "log_lvl"    : "${RADICAL_DEFAULT_LOG_LVL:ERROR}",
     "log_tgt"    : "${RADICAL_DEFAULT_LOG_TGT:.}",
     "log_dir"    : "${RADICAL_DEFAULT_LOG_DIR:$PWD}",
-    "report"     : "${RADICAL_DEFAULT_REPORT:TRUE}",
+    "report"     : "${RADICAL_DEFAULT_REPORT:True}",
```
mtitov, line 6: keep it with capital letters - `TRUE`?

src/radical/utils/reporter.py
```diff
@@ -106,16 +106,23 @@ def __init__(self, name, ns=None, path=None, targets=None):
...
-            self._enabled = str(ru_get_env_ns('report', ns)).lower()
+        if enabled is not None:
+            self._enabled = bool(enabled)
+        else:
+            self._enabled = str(ru_def['report']).lower()
+
+            if ru_get_env_ns('report', ns) is not None:
+                self._enabled = str(ru_get_env_ns('report', ns)).lower()
+
+            if self._enabled in ['0', 'false', 'off', False, None]:
```
mtitov, line 117: let's keep just `['0', 'false', 'off', 'none']`? since we will have strings only

mtitov (approved):
LGTM! Thanks, Andre!

p.s. I've added couple comments just for a small cleanup, you can skip them if you want

src/radical/utils/configs/utils_default.json
andre-merzky: Thanks, fixed.

src/radical/utils/reporter.py
andre-merzky: Ack, fixed - good catch.

PR #400, opened 2024-03-16
add flux module path

andre-merzky:
This fixes radical-cybertools/radical.pilot/issues/3149


mtitov (approved):
LGTM!

p.s. tested it without setting PYTHONPATH on my side, python flux was discovered correctly

PR #401, opened 2024-04-05
implement custom (de)serialization for json and msgpack

andre-merzky:


src/radical/utils/serialize.py
```diff
@@ -0,0 +1,179 @@
...
+def _prep_typed_dict(d):
+    from .typeddict import TypedDict
+    tmp = dict()
+    for k,v in d.items():
+        if isinstance(v, TypedDict):
+            tmp[k] = _prep_typed_dict(v)
+        else:
+            tmp[k] = v
+    tmp['_xtype'] = type(d).__name__
+    return tmp
```
mtitov, line 48: quick check - this function is what we have as a corresponding method `d.as_dist()` (if `d` is `TypedDict`) as well we have it as a function `from typeddict import as_dict`

(*) only the difference is you would need to add `_xtype` key-value separately

did I understand this func correctly?

andre-merzky: Basically yes, but a but less sophisticated than `as_dict` (no consideration for lists of typed dicts for example).

andre-merzky: I added the type annotation to `as_dict` now so that code can be reused...

andre-merzky:
I iterated more on this, the PR should be ready for another review.  Thanks!

src/radical/utils/typeddict.py
```diff
@@ -276,7 +280,7 @@ def __getattr__(self, k):
                 self.__dict__['_data'] = dict()
             return self.__dict__['_data']
 
-        if k.startswith('__'):
+        if k.startswith('_'):
             return object.__getattribute__(self, k)
```
mtitov, line 284: Isn't this will break the use of attributes with the underscore?
For example we have underscore in `ru.Message` types

src/radical/utils/typeddict.py
```diff
@@ -288,15 +292,15 @@ def __getattr__(self, k):
 
     def __setattr__(self, k, v):
 
-        # if k.startswith('_'):
-        #     return object.__setattr__(self, k, v)
+        if k.startswith('_'):
+            return object.__setattr__(self, k, v)
```
mtitov, line 296: same regarding using `_attr` attributes

src/radical/utils/typeddict.py
```diff
@@ -288,15 +292,15 @@ def __getattr__(self, k):
...
 
         self._data[k] = self._verify_setter(k, v)
 
     def __delattr__(self, k):
 
-        # if k.startswith('_'):
-        #     return object.__delattr__(self, k)
+        if k.startswith('_'):
+            return object.__delattr__(self, k)
```
mtitov, line 303: same regarding using `_attr` attributes

src/radical/utils/json_io.py
```diff
@@ -86,7 +98,7 @@ def parse_json(json_str, filter_comments=True):
             content += re.sub(r'^\s*#.*$', '', line)
             content += '\n'
 
-        return json.loads(content)
+        return from_json(content)
```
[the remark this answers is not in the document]
andre-merzky: Right!  I simplified that bit now.

src/radical/utils/typeddict.py
andre-merzky: Thanks for the note, let me check this.

src/radical/utils/typeddict.py
```diff
@@ -138,6 +138,10 @@ def __init__(self, from_dict=None, **kwargs):
               `kwargs`).
         '''
 
+        from .serialize import register_serializable
+
+        register_serializable(self.__class__)
```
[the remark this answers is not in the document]
mtitov: call `register_serializable` only when class is imported and not every time instance is created
(also it is related to the class and not to the instance, and since we have meta class it should be there)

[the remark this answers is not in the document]
andre-merzky: We are running into circular imports when doing too many things on `import` - as is the case with the above proposal.  Yes, it can probably be solved, but I rather keep the code simple as is.  We should measure before optimizing on that level...

src/radical/utils/typeddict.py
andre-merzky: I reverted that now - thanks for catching!

src/radical/utils/typeddict.py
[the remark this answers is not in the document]
mtitov: I see the issue, and would propose the following the solution

(1) in `typeddict` module:
```python
        # within TypedDictMeta.__new__
        new_cls = super().__new__(mcs, name, bases, namespace)
        if new_cls.__base__ is not dict:
            # register sub-classes
            from .serialize import register_serializable
            register_serializable(new_cls)
        return new_cls
``` 
(2) in `serialize` module
```python
from .typeddict import as_dict, TypedDict
# after register_serializable is defined
register_serializable(TypedDict)
```
p.s., `TypedDict` is an internal type, thus it is fine if we'll register it manually

> We should measure before optimizing on that level...

sorry, disagree, this is not an optimization - you introduce a new functionality with the minimal impact on the existing structures  - and code above minimizes the impact, `TypedDict` is base for messages and tasks, so we should avoid adding code into initialization block as much as possible

andre-merzky:
Thanks @mtitov, applied as suggested.

mtitov (approved):
LGTM! Thanks!

PR #407, opened 2024-05-13
updated `expr` usage for env funcs extraction

mtitov:


andre-merzky (approved):
Different bash versions?  Oh the joy...

PR #408, opened 2024-05-28
Fix/flux remote

andre-merzky:


src/radical/utils/flux.py
```diff
@@ -144,7 +144,10 @@ def _locked_start_service(self,
         if launcher:
             cmd += shlex.split(launcher)
 
-        cmd += ['flux', 'start', 'bash', '-c', 'echo URI:$FLUX_URI && sleep inf']
+        cmd += ['flux', 'start', 'bash', '-c',
+                'echo "HOST:$(hostname) URI:$FLUX_URI" && sleep inf']
+
+        self._log.debug('flux command', cmd)
```
mtitov, line 150: missing reference to the cmd in the msg -> `'flux command: %s'`

src/radical/utils/flux.py
```diff
@@ -190,6 +199,8 @@ def _locked_start_service(self,
         self._env       = flux_env
         self._proc      = flux_proc
 
+        self._log.debug('flux uri', flux_uri)
```
mtitov, line 202: `'flux uri: %s'`

mtitov (approved):
LGTM! couple comments regarding debug messages

src/radical/utils/flux.py
andre-merzky: thanks!

src/radical/utils/flux.py
andre-merzky: Thanks also, both fixed.

PR #410, opened 2024-06-27
apply blacklist

andre-merzky:
This is in support of radical.pilot/issues/3183

### DIFF
--- a/src/radical/utils/env.py
+++ b/src/radical/utils/env.py
@@ -59,7 +59,7 @@ def env_read(fname: str) -> Dict[str, str]:
 
 # ------------------------------------------------------------------------------
 #
-def env_write(script_path, env, unset=None, pre_exec=None):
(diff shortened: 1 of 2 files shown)


src/radical/utils/env.py
```diff
@@ -78,11 +78,25 @@ def env_write(script_path, env, unset=None, pre_exec=None):
             data += 'unset %s\n' % k
         data += '\n'
 
+        env = {k: v for k, v in env.items() if k not in BLACKLIST}
+
+    if blacklist:
```
mtitov, line 83: there is a confusion between `BLACKLIST` (which env variables are `unset`) and `blacklist` (which env variables are skipped)

Besides this all looks good

andre-merzky: Thanks, I renamed `BLACKLIST` to `IGNORE_LIST` now.

PR #415, opened 2024-08-26
remove an obsolete monkeypatch

andre-merzky:


andre-merzky:
I am not sure why the tests fail here - they seem to work alright alright.  Any insight?

.github/workflows/ci.yml
```diff
@@ -43,7 +43,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11' ]
+        python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11', '3.12',
+                          '3.13.0-rc.1' ]
```
mtitov, line 47: seems that rc2 is available

mtitov: on another note - should we skip 3.13 until coverage is fixed there?
OR we can have a draft PR with 3.13, and check it every sprint until it is passed

mtitov (approved):
LGTM!

PR #420, opened 2024-10-15
Feature/simple hb

andre-merzky:
Implement a much simpler process watcher than what is used in RP atm.  This will support the airflow executor in ScaleMS.


src/radical/utils/heartbeat.py
```diff
@@ -263,5 +264,136 @@ def wait_startup(self, uids=None, timeout=None):
...
+    #
+    def _watch(self):
+
+        self._log.debug_1('pwatcher started')
+
+        while True:
+
+            with self._lock:
+
+                for pid in self._pids:
```
mtitov, line 330: should be `for pid in list(self._pids):`, otherwise it will not `remove` from the list it going through

src/radical/utils/heartbeat.py
```diff
@@ -263,5 +264,136 @@ def wait_startup(self, uids=None, timeout=None):
...
+
+
+    # --------------------------------------------------------------------------
+    #
+    def unwatch(self, pid):
+
+        self._log.debug('remove pid %d from watchlist', pid)
+
+        with self._lock:
+            self._pids.remove(pid)
```
mtitov, line 361: just in case better to check that `pid` is in `self._pids` and wasn't terminated b/c of being not alive

src/radical/utils/heartbeat.py
```diff
@@ -263,5 +264,136 @@ def wait_startup(self, uids=None, timeout=None):
...
+
+    # --------------------------------------------------------------------------
+    #
+    def _killall(self, pid):
+
+        self._log.debug("process %d's demise triggered killall (%s)",
+                        pid, self._pids)
+
+        for pid in self._pids:
+            os.kill(pid, signal.SIGKILL)
```
mtitov, line 387: `pid` process might not be in `self._pids` (see L335)

src/radical/utils/heartbeat.py
andre-merzky: Thanks!

src/radical/utils/heartbeat.py
andre-merzky: ack, done.

src/radical/utils/heartbeat.py
andre-merzky: and it might be dead - shielded now by `try/except`.  Thanks.

mtitov (approved):
LGTM! Thanks!

PR #421, opened 2024-10-21
Feature/log delay

andre-merzky:
This PR delays the creation of log- and profiles until the first write, thus avoiding empty files.

It is strange: I am quite positive that we implemented that in the past already, but somehow seem to have lost that code?  Oh well, fortunately the fix is simple, and here it is again...


mtitov (approved):
LGTM!

PR #423, opened 2024-10-23
Fix/client log

andre-merzky:
This supports https://github.com/radical-cybertools/radical.pilot/pull/3244 by passing a logger to a zmq client.


mtitov (approved):
LGTM!

PR #427, opened 2024-11-24
Feature/find port

andre-merzky:
when auto-assigning port numbers, use 10,000 as starting point to scan for free ports.


mtitov:
@andre-merzky confirming that this was tested in RP CI runs

PR #429, opened 2024-12-03
zmq startup is faster than we think

andre-merzky:
This shaves a couple of seconds from session and component startup in RP.


AymenFJA (approved):
LGTM, thanks Andre.

PR #432, opened 2025-02-05
add comment on process group killing

andre-merzky:
This PR is to discuss if (a) we need the PG kill as fallback, and if so, (b) how to ensure that pytest is not affected.


andre-merzky:
Closing in favor of #433

PR #433, opened 2025-02-20
this fixes the test aborts

andre-merzky:
pytest does not like `SIGKILL` it turns out.  This PR replaces #432

### DIFF
--- a/src/radical/utils/heartbeat.py
+++ b/src/radical/utils/heartbeat.py
@@ -322,20 +322,18 @@ class PWatcher(object):
         try   : os.killpg(pid, signal.SIGTERM)
         except: pass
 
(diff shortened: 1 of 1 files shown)


AymenFJA (approved):
LGTM. Thanks, Andre.

mtitov:
@andre-merzky seems that the issue with tests is not resolved https://github.com/radical-cybertools/radical.pilot/actions/runs/13451583777/job/37608735061

andre-merzky:
Ah dammit - I was finally able to reproduce that locally, and the commit *did* resolve it in that case...  But now I am unsure how to debug further...  Hmmm....  Thanks for letting me know!

PR #434, opened 2025-02-27
remove deepcopy, allow `as_dict` overwrite

andre-merzky:
This changes the semantics of `ru.TypedDict` to get back in sync with the default `dict` semantic, and specifically that non-trivial elements are stored by reference, not by value.  At the same time this improves performance by about 50%.

A second set of changes to the `as_dict` method now allows to overwrite that method for classes inheriting from `TypeDict`.

### DIFF
--- a/src/radical/utils/typeddict.py
+++ b/src/radical/utils/typeddict.py
@@ -77,17 +77,22 @@ class TypedDictMeta(type):
             '_defaults'    : {},
             '_self_default': False,  # convert unschemed-dict into class itself
             '_check'       : False,  # attribute type checking on set
-            '_cast'        : True    # attempt to cast on type mismatch
+            '_cast'        : True,   # attempt to cast on type mismatch
+            '_deep'        : True    # use deepcopy on defaults
         }
 
         for _cls in bases:
+            deep = namespace.get('_deep', _base_namespace['_deep'])
             for k in _base_namespace.keys():
                 _cls_v = getattr(_cls, k, None)
                 if _cls_v is not None:
                     if   k == '_schema':
                         _base_namespace[k].update(_cls_v)
                     elif k == '_defaults':
-                        _base_namespace[k].update(_cls_v)
+                        if deep:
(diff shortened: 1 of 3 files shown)


src/radical/utils/typeddict.py
```diff
@@ -151,7 +151,7 @@ def __init__(self, from_dict=None, **kwargs):
 
         register_serializable(self.__class__)
 
-        self.update(copy.deepcopy(self._defaults))
+        self.update(self._defaults)
```
mtitov, line 154: with this we'll have reference issue as mentioned in the description
```python
class TDSimple(TypedDict):

    _schema = {
        'attr_dict': {str: None}
    }

    _defaults = {
        'attr_dict': {'dict_key_00': 100}
    }

tds_1 = TDSimple()
tds_2 = TDSimple()

assert tds_1.as_dict() == tds_2.as_dict()
tds_1.attr_dict['new_key'] = 'new_value'
assert tds_1.as_dict() == tds_2.as_dict()
```
`tds_2` will have its attribute to be changed

andre-merzky: that is exactly the point - this *is* the expected behavior for dictionaries!

```py
common = {'dict_key_00': 100}
ds_1 = {'attr_dict': common}
ds_2 = {'attr_dict': common}

assert ds_1 == ds_2
ds_1['attr_dict']['new_key'] = 'new_value'
assert ds_1 == ds_2
```

I understand that this was used differently in your code (for convenience IIRC), but that does not justify (a) breaking expectations, and (b) 50% performance hit (this really is on our critical path in all applications).  The `deepcopy` needs to happen here:
```
ds_1 = {'attr_dict': copy.deepcopy(common)}
ds_2 = {'attr_dict': copy.deepcopy(common)}
```

This implies revisiting the assignment of list or dict instance as `default values` - those should get then deepcopied when creating instances.

mtitov (approved):
LGTM!

PR #440, opened 2025-03-12
atomic write_json

### LINKED ISSUE #439: Enhance json-write method
Use tmp file as an intermediate step, not to overwrite existing file directly

```python
def write_json(fname, data):
    if not data: return
    with tmpfile as fout:
        fout.write(json.dumps(data.as_dict()))
    os.rename(tmpfile.name, fname)
```

andre-merzky:
This fixes #439 439

### DIFF
--- a/src/radical/utils/json_io.py
+++ b/src/radical/utils/json_io.py
@@ -6,6 +6,8 @@ __license__   = 'MIT'
 
 
 import re
+import os
+import tempfile
 
 from .serialize import to_json, from_json
 from .misc      import as_string, ru_open
@@ -51,8 +53,11 @@ def read_json_str(filename, filter_comments=True):
 def write_json(data, fname):
     '''
     thin wrapper around python's json write, for consistency of interface
-
     '''
+
+    if not data:
+        # we don't have a logger to report :-/
+        return
 
     if isinstance(fname, dict) and isinstance(data, str):
         # arguments were switched: accept anyway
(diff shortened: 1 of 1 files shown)


src/radical/utils/json_io.py
```diff
@@ -51,19 +53,27 @@ def read_json_str(filename, filter_comments=True):
 def write_json(data, fname):
     '''
     thin wrapper around python's json write, for consistency of interface
-
     '''
 
+    if not data:
+        # we don't have a logger to report :-/
+        return
```
mtitov, line 60: let's move it right before openning the file and add `to_json` to it
```python
data = to_json(data)
if not data:
    return
```

src/radical/utils/json_io.py
```diff
@@ -51,19 +53,27 @@ def read_json_str(filename, filter_comments=True):
...
-        f.write(to_json(data))
-        f.write('\n')
-        f.flush()
+    dirname = os.path.dirname(fname) or '.'
+
+    _, t_name = tempfile.mkstemp(dir=dirname)
+    with open(t_name, 'w') as f_out:
+        f_out.write(to_json(data))
+        f_out.write('\n')
+        f_out.close()
```
mtitov, line 74: we don't need `close()` here since `with open` takes care of it, and we can keep just one `write` operation (`f_out.write(data + '\n'`)

src/radical/utils/json_io.py
andre-merzky: Ack on moving to after the arg check, I missed that - but the `to_json` call won't  do anything on an empty dict or list  - what am I missing?

src/radical/utils/json_io.py
andre-merzky: ack, fixed.

src/radical/utils/json_io.py
mtitov: with that I thought to avoid situation when `to_json` may "corrupt" the data, i.e., to ensure that `to_json` returns non empty string

src/radical/utils/json_io.py
mtitov: and let's keep a single `write` operation

andre-merzky: Hmm, that's quite a micro-optimization - what's the purpose?  That call is not in a critical path.  But patch is applied as suggested.

src/radical/utils/json_io.py
mtitov: @andre-merzky this one missed

mtitov (approved):
LGTM! Thank you!

PR #442, opened 2025-03-31
move python version to 3.8

andre-merzky:


bin/radical-stack-clone
```diff
@@ -200,7 +200,7 @@ then
     then
         if test -z "$python"
         then
-            conda create -y -p "$veloc" python=3.7 >> $log 2>&1
+            conda create -y -p "$veloc" python=3.8 >> $log 2>&1
```
mtitov, line 203: not sure if we want to have here at least 3.9, but that just as a suggestion, not a stoping point

mtitov (approved):
LGTM! (and one comment)

@andre-merzky

Issue #443, opened 2025-04-01
Purude Anvil Network Interface requires infiniband


AymenFJA:
Anvil network interface requires enabling `infiniband`.

PR #444, opened 2025-04-01
Enable Anvil Infinitband interface

AymenFJA:
addresses #443

### DIFF
--- a/src/radical/utils/host.py
+++ b/src/radical/utils/host.py
@@ -56,7 +56,7 @@ def get_hostip(req=None, log=None):
 
     white_list = [
         'ens1f1',   # amarel
-      # 'ib0',      # infiniband
+        'ib0',      # infiniband
         'hsn0',     # Frontier (HPE Cray EX)
         'ipogif0',  # Cray's
         'br0',      # SuperMIC


src/radical/utils/host.py
```diff
@@ -56,7 +56,7 @@ def get_hostip(req=None, log=None):
 
     white_list = [
         'ens1f1',   # amarel
-      # 'ib0',      # infiniband
+        'ib0',      # infiniband
```
mtitov, line 59: please add comment here that it is for Anvil, and that it doesn't fit Amarel (based on comments for previous commits here). Eventially these interfaces will be moved into resource configuration files

andre-merzky:
I added that comment - but indeed, that might break Amarel.  I opened #445 for a cleaner solution.  Either way, I leave it to your judgement is this should be merged (i.e., what machine you want to break :-P )

andre-merzky:
Now also fixes #445

src/radical/utils/host.py
```diff
@@ -52,11 +51,14 @@ def get_hostip(req=None, log=None):
     # Then this list is traversed, we check if the interface exists and has an
     # IP address.  The first match is used.
 
+    if not req:
+        req = DefaultConfig().get('iface')
```
mtitov, line 55: could a corresponding env variable be checked here directly? for me it doesn't seem fit into `DefaultConfig`, which is about logging/reporting/profiling

and if env variable is set it can expand `req` (i.e., even if `req` is provided, still to add the value of `RADICAL_IFACE`)
```
req = as_list(req) + as_list(os.getenv('RADICAL_IFACE'))
```

src/radical/utils/host.py
```diff
@@ -52,11 +51,14 @@ def get_hostip(req=None, log=None):
...
+
     req = as_list(req)
 
     white_list = [
         'ens1f1',   # amarel
-      # 'ib0',      # infiniband
+        'ib0',      # infiniband   # NOTE: unusable on Amarel, use `req` there!
```
mtitov, line 61: \+ add comment that it is for anvil

src/radical/utils/host.py
```diff
@@ -69,6 +71,10 @@ def get_hostip(req=None, log=None):
         'sit0'      # ?
     ]
 
+    # default value is cached
+    if not req and _hostip:
+        return _hostip
```
mtitov, line 76: that's tricky about caching - could it be that the first call is with `req` and the following one is without?

mtitov: caching could be with explicit association of `req` and `_hostip` (set `_hostip` as `dict` with the corresponding `req` as a key)

src/radical/utils/host.py
```diff
@@ -104,9 +110,10 @@ def get_hostip(req=None, log=None):
         if log:
             log.debug('check iface %s: ip is %s', iface, ip)
 
-        if ip:
+        if not req and ip:
             _hostip = ip
-            return ip
+
+        return ip
```
mtitov, line 116: this was under `if ip` statement, thus not to return `None`, should it be as following?
```python
       if ip:
           if not req:
               _hostip = ip
           return ip
```

btw, here you want to cache only "default" `ip`?

src/radical/utils/host.py
mtitov: and also, for amarel the first one (`'ens1f1'`) should be picked, so nit sure that `req` is require for amarel

src/radical/utils/host.py
andre-merzky: That would not help RP: Environment should ideally be set before process startup, but RP reads the interface name from the config file runtime.  Also, given that we *hardode* a default list, I think that the default config is a perfect place to adjust that :-)

src/radical/utils/host.py
andre-merzky: Oh, then I misunderstood your request above.  Why was `ib0` disabled then?  But either way, I think that comment should better go to the amarel / anvil resource configs, and RU should stay clear of machine specific settings as much as possible.  Would that work for you?

src/radical/utils/host.py
andre-merzky: Yes, or with a different `req`, and that's why I wanted to disable caching for the non-default lookups.

src/radical/utils/host.py
andre-merzky: Yes.

src/radical/utils/host.py
andre-merzky: I also should mention that `RADICAL_DEFAULT_IFACE` is evaluated when the config file is read.

AymenFJA:
@andre-merzky @mtitov, seeing that this PR is required by our IMPRESS runs on Anvil, should we consider this done and ready to be merged?

andre-merzky:
TODO AM: remove default config entry for `iface`

andre-merzky:
@AymenFJA : @mtitov will go over this topic once more and we'll make a decision early next week.

src/radical/utils/host.py
mtitov: keeping comments here before adding corresponding parameters (`iface`) within resource configurations

mtitov (approved):
LGTM!

p.s., Andre convinced me regarding this approach - `DefaultConfig` is a Singleton, thus its instance is created once, and it allows to share it between layers (RP and RU) - creating an instance within RP layer with the corresponding `iface`, then will be available within RU at `get_hostip` method

src/radical/utils/host.py
```diff
@@ -104,9 +101,7 @@ def get_hostip(req=None, log=None):
         if log:
             log.debug('check iface %s: ip is %s', iface, ip)
 
-        if ip:
-            _hostip = ip
-            return ip
+        return ip
```
mtitov, line 104: @andre-merzky one last thing here - should you check if `ip` has any value?
```python
if ip:
    return ip
```

andre-merzky: Ack, done.

mtitov:
@andre-merzky ping

PR #446, opened 2025-04-15
stable service startup and port selection

andre-merzky:
This simplifies `xGFabric` deplyment, and service deployment in general


src/radical/utils/zmq/server.py
```diff
@@ -249,21 +250,10 @@ def _work(self) -> None:
...
-                break
-            except zmq.error.ZMQError as e:
-                if 'Address already in use' in str(e):
-                    self._log.warn('port in use - try next (%s)' % url)
-                else:
-                    raise
-
-        addr       = Url(as_string(self._sock.getsockopt(zmq.LAST_ENDPOINT)))
-        addr.host  = get_hostip()
+        addr = zmq_bind(self._sock)
```
mtitov, line 253: this disregards provided url (`proto`, `iface`, and `ports`)

p.s., would be good to combine/merge `misc.find_port` with `zmq.server.Server._iterate_ports` and then extend `zmq.utils.zmq_bind` to handle provided `url` as it is done for `Server`

src/radical/utils/misc.py
```diff
@@ -830,13 +830,21 @@ def ru_open(*args, **kwargs):
 
 # ------------------------------------------------------------------------------
 #
-def find_port(port_min=10000, port_max=65535):
+def find_port(port_min: int = None,
+              port_max: int = None) -> Union[int, None]:
```
andre-merzky, line 834: Add `iface`?

